PR #3, opened 2023-06-23
Np 44837 index nvi candidate

monaullah:
https://unit.atlassian.net/browse/NP-44837

IndexNviCandidateHandler consumes messages from the SQS queue _NewCandidateQueue_, reads the expanded resource from the S3 bucket _resources_, creates an nvi candidate index document and indexes the document in a new index: _nvi-candidates_.

Generation of the nvi candidate index document from the expanded resource (in S3 bucket _resources_) is done with json pointers for now, but the plan is to refactor this and use sparql to do this for a more generic solution (new PR to come).

(Sorry for the big PR)


index-handlers/src/main/java/no/sikt/nva/nvi/index/aws/OpenSearchIndexClient.java
```diff
@@ -0,0 +1,82 @@
...
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.http.apache.ApacheHttpClient;
+import software.amazon.awssdk.regions.Region;
+
+@JacocoGenerated
+//TODO: Handle test coverage
+public class OpenSearchIndexClient implements IndexClient<NviCandidateIndexDocument> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(OpenSearchIndexClient.class);
+    private static final String INDEX = "nviCandidates";
```
sondrev, line 27: nvi-candidates
The index name will be used in URL paths, and kebab-case should be more suited

index-handlers/src/main/java/no/sikt/nva/nvi/index/aws/OpenSearchIndexClient.java
```diff
@@ -0,0 +1,82 @@
...
+                   .document(indexDocument)
+                   .build();
+    }
+
+    private static AwsSdk2TransportOptions transportOptionWithToken(String token) {
+        return AwsSdk2TransportOptions.builder().addHeader(AUTHORIZATION, token).build();
+    }
+
+    private boolean indexExists() throws IOException {
+        try {
```
sondrev, line 63: use attempt() instead
Looks better, and its used in other nva projects in place of direct try/catch.

index-handlers/src/main/java/no/sikt/nva/nvi/index/aws/OpenSearchIndexClient.java
```diff
@@ -0,0 +1,82 @@
...
+                       .exists(ExistsRequest.of(s -> s.index(INDEX)))
+                       .value();
+        } catch (IOException | OpenSearchException e) {
+            LOGGER.error("Error while checking if index {} exists. Error: {}", INDEX, e.getMessage());
+            throw e;
+        }
+    }
+
+    private void createIndex() throws IOException {
+        try {
```
sondrev, line 74: use attempt() instead
Looks better, and its used in other nva projects in place of direct try/catch.

index-handlers/src/main/java/no/sikt/nva/nvi/index/aws/OpenSearchIndexClient.java
```diff
@@ -0,0 +1,82 @@
...
+    private final OpenSearchClient openSearchClient;
+
+    public OpenSearchIndexClient(String openSearchEndpoint, CachedJwtProvider cachedJwtProvider, Region region)
+        throws IOException {
+        this.openSearchClient = new OpenSearchClient(
+            new AwsSdk2Transport(
+                ApacheHttpClient.builder().build(),
+                openSearchEndpoint,
+                region,
+                transportOptionWithToken("Bearer " + cachedJwtProvider.getValue().getToken())
```
sondrev, line 37: The token you are providing will only last 15mins, but if your lambda is called repeatedly it can survive longer and the token will be invalid.
Instead you need to refresh the token on each call if its outdated. See search-api

index-handlers/src/main/java/no/sikt/nva/nvi/index/model/Publication.java
```diff
@@ -0,0 +1,87 @@
+package no.sikt.nva.nvi.index.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import java.util.Objects;
+import nva.commons.core.JacocoGenerated;
+
+public final class Publication {
```
sondrev, line 8: Can this be called something else so its not condused with Publication from the datamodel?

sondrev (changes requested):

index-handlers/src/main/java/no/sikt/nva/nvi/index/model/Publication.java
monaullah: Renamed to PublicationDetails

index-handlers/src/main/java/no/sikt/nva/nvi/index/utils/NviCandidateIndexDocumentGenerator.java
```diff
@@ -0,0 +1,144 @@
...
+    private static String extractYear(JsonNode resource) {
+        return resource.at(JSON_PTR_PUBLICATION_DATE_YEAR).textValue();
+    }
+
+    private static String formatPublicationDate(JsonNode publicationDateNode) {
+        var year = publicationDateNode.get(FIELD_PUBLICATION_DATE_YEAR);
+        var month = publicationDateNode.get(FIELD_PUBLICATION_DATE_MONTH);
+        var day = publicationDateNode.get(FIELD_PUBLICATION_DATE_DAY);
+
+        return attempt(() -> LocalDate.of(year.asInt(), month.asInt(), day.asInt()).toString()).orElse(
```
brinxmat, line 141: The safest way of doing this — I think — is treating it as an ISO 8601 string with simple concatenation c.f. https://www.w3.org/TR/NOTE-datetime#:~:text=options%20it%20permits.-,Formats,-Different%20standards%20may

Hence: 2000, 2000-01, 2000-01-01

The trouble is that these are strings and that's all — it's deliberate. We observe a some weird stuff in these fields on import. We also assume Gregorian dates which will bite us at some point.

nvi-commons/src/main/java/no/sikt/nva/nvi/common/Something.java
brinxmat, line 1: Something has been removed here…

nvi-commons/src/test/java/no/sikt/nva/nvi/common/SomethingTest.java
brinxmat, line 1: Something else has been…

brinxmat (changes requested):

index-handlers/src/main/java/no/sikt/nva/nvi/index/aws/OpenSearchIndexClient.java
monaullah: As discussed, solving in new PR

brinxmat (approved):

PR #16, opened 2023-07-07
Np 44995 remove candidate from index

truhacevkir:
Could be great to remove @JacocGenerated from OpenSearchClient, but for now it is not possible to test indexExists() method when throwing exception. 

I have also removed unused environment variables in template.

One more thing, it seems like OpenSearch is smart enough and does not create new document when document with matching id already exists. Which means we do not need to update(or delete and put new one) nvi candidate in index when update event from expansion comes.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -21,10 +21,6 @@ Globals:
       AllowOrigin: "'*'"
 
 Parameters:
-  CognitoAuthorizerArn:
-    Type: 'AWS::SSM::Parameter::Value<String>'
-    Description: Reference to Cognito UserPool for the stage
-    Default: CognitoAuthorizerArn
   CognitoAuthorizationUri:
     Type: 'AWS::SSM::Parameter::Value<String>'
     Default: '/NVA/CognitoUri'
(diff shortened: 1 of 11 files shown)


template.yaml
```diff
@@ -227,12 +227,27 @@ Resources:
...
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: index-handlers
+      Handler: no.sikt.nva.nvi.index.DeleteNviCandidateHandler::handleRequest
+      Timeout: 30
+      MemorySize: 1536
+      Role: !GetAtt NvaNviRole.Arn
+      Environment:
+        Variables:
+          SEARCH_INFRASTRUCTURE_API_HOST: !Ref SearchInfrastructureApiHost
```
monaullah, line 248: Don't think it is necessary to set these variables here since they are global environment variables.

monaullah: You could also remove them as global env variables and only define them in the handlers that use them

ketilaa (approved):

PR #17, opened 2023-07-10
Np 44997 facet for approval status

truhacevkir:
We get following aggregations:

```
{
	"approvalStatus.rejected": {
		"docCount": 1
	},
	"approvalStatus.pending": {
		"docCount": 2
	},
	"approvalStatus.approved": {
		"docCount": 1
	}
}
```

### DIFF
--- a/index-handlers/src/main/java/no/sikt/nva/nvi/index/Aggregations.java
+++ b/index-handlers/src/main/java/no/sikt/nva/nvi/index/Aggregations.java
@@ -1,0 +1,66 @@
+package no.sikt.nva.nvi.index;
+
+import java.util.Map;
+import no.sikt.nva.nvi.index.model.ApprovalStatus;
+import org.opensearch.client.opensearch._types.aggregations.Aggregation;
+import org.opensearch.client.opensearch._types.query_dsl.BoolQuery.Builder;
+import org.opensearch.client.opensearch._types.query_dsl.Query;
+import org.opensearch.client.opensearch._types.query_dsl.QueryStringQuery;
+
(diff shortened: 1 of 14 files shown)


index-handlers/src/main/java/no/sikt/nva/nvi/index/Aggregations.java
```diff
@@ -0,0 +1,66 @@
...
+    private static Query queryToMatchApprovalStatus(ApprovalStatus approvalStatus) {
+        return new Query(new QueryStringQuery.Builder()
+                             .query(approvalStatus.toString())
+                             .defaultField(jsonPathOf(AFFILIATIONS, APPROVAL_STATUS))
+                             .build());
+    }
+
+    private static Map<String, Aggregation> constructAggregations() {
+        return Map.of(
+            jsonPathOf(APPROVAL_STATUS, PENDING), pendingAggregation(),
```
ketilaa, line 61: Do we really have to explicitly mention all specific values to aggregate over?

ketilaa (approved):

index-handlers/src/main/java/no/sikt/nva/nvi/index/Aggregations.java
truhacevkir: In this case yes, since we have do not have single status for each index document, but list of values for every affiliation. So to construct this aggregation I had actually to "filter out" statuses that are not relevant.
For example approvedAggregation() is aggregation for documents that do not contain statuses PENDING and REJECTED

PR #20, opened 2023-08-09
Np 45083 upsert nvi candidate handler validation

monaullah:
Splitting Jira task (https://unit.atlassian.net/browse/NP-45083) into several PRs to keep them small. This is nr 1
- UpsertNviCandidateHandler consumes messages on sqs queue "NewCandidateQueue"
- Logs error if message body is invalid
- Logs error if publicationBucketUri is null

### DIFF
--- a/event-handlers/src/main/java/handlers/UpsertNviCandidateHandler.java
+++ b/event-handlers/src/main/java/handlers/UpsertNviCandidateHandler.java
@@ -16,8 +16,6 @@ public class UpsertNviCandidateHandler implements RequestHandler<SQSEvent, Void>
 
     private static final Logger LOGGER = LoggerFactory.getLogger(UpsertNviCandidateHandler.class);
 
-    private static final String ERROR_MESSAGE_BODY_INVALID = "Message body invalid";
-
     @Override
     public Void handleRequest(SQSEvent input, Context context) {
         input.getRecords()
@@ -54,6 +52,6 @@ public class UpsertNviCandidateHandler implements RequestHandler<SQSEvent, Void>
     }
 
     private void logInvalidMessageBody(String body) {
-        LOGGER.error(ERROR_MESSAGE_BODY_INVALID, body);
+        LOGGER.error("Message body invalid: {}", body);
     }
 }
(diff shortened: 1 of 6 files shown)


event-handlers/src/main/java/handlers/UpsertNviCandidateHandler.java
```diff
@@ -0,0 +1,59 @@
...
+                       return null;
+                   });
+    }
+
+    //TODO: Remove jacocoGenerated when "happy cases" are implemented
+    @JacocoGenerated
+    private UpsertRequest validate(UpsertRequest request) {
+        if (isNull(request.publicationBucketUri())) {
+            logInvalidMessageBody(request.toString());
+            return null;
```
truhacevkir, line 49: I guess sending null to upsertNviCandidate method will cause exception, maybe throw exception already here instead of sending null to next method.

truhacevkir (changes requested):

event-handlers/src/main/java/handlers/UpsertNviCandidateHandler.java
monaullah: The reason for not throwing an exception here is that SQSEvent input contains _multiple_ records. Therefore, if the message body of one record is invalid, returning null will make sure the lambda processes the _next_ record. If we throw an exception here, the next records will not be processed.

monaullah: https://docs.aws.amazon.com/lambda/latest/dg/with-sqs-create-package.html#with-sqs-example-deployment-pkg-java

truhacevkir: Okey, this handler will recieve sqs event from evaluate handler and as I see evaluate handler sends event with single record. But you are probably right.

truhacevkir:
You can create interface for database service which in feature will be extended by nviService or something like that, so you can create fakeNviService and test with it, so the only remaining thing to do when nviService is implemented is to substitute it with real nviService.

truhacevkir (approved):

PR #28, opened 2023-08-17
Upsert nvi candidate, add publication details

monaullah:
Splitting Jira task (https://unit.atlassian.net/browse/NP-45083) into several PRs to keep them small. This is nr 3

UpsertNviCandidateHandler consumes messages on sqs queue "NewCandidateQueue". If the candidate does not already exist in DB, create new candidate with pending approval statuses and save.
Faking NviCandidateRepository until DB Tech is chosen and implemented

In this PR: Updated message body that is consumed by UpsertNviCandidateHandler and add candidate details to Candidate when saving


nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/NviService.java
```diff
@@ -38,4 +63,30 @@ private static List<ApprovalStatus> generatePendingApprovalStatuses(List<URI> in
...
+                                   .withPublicationId(candidateDetails.publicationId())
+                                   .withIsApplicable(true)
+                                   .withCreators(mapToVerifiedCreators(candidateDetails.verifiedCreatorDtos()))
+                                   .withLevel(Level.parse(candidateDetails.level()))
+                                   .withInstanceType(candidateDetails.instanceType())
+                                   .withPublicationDate(mapToPublicationDate(publicationDate))
+                                   .withApprovalStatuses(
+                                       generatePendingApprovalStatuses(extractInstitutionIds(candidateDetails)))
+                                   .build();
+        nviCandidateRepository.save(pendingCandidate);
```
truhacevkir, line 84: extract to method, toPendingCandidate()

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/NviService.java
```diff
@@ -2,34 +2,59 @@
...
+        }
     }
 
-    public boolean exists(URI publicationId) {
-        return nviCandidateRepository.findByPublicationId(publicationId.toString()).isPresent();
+    private static List<VerifiedCreator> mapToVerifiedCreators(List<VerifiedCreatorDto> verifiedCreatorDtos) {
+        return verifiedCreatorDtos.stream()
+                   .map(verifiedCreatorDto -> new VerifiedCreator(
+                       verifiedCreatorDto.id(), verifiedCreatorDto.nviInstitutions()))
+                   .toList();
```
truhacevkir, line 39: you can create method toDao() for your dtos, for example candidate.toDao()

truhacevkir: The same for dao, candidateDao.toDto() or toCandidate(), I guess sondre will need these methods

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/NviService.java
truhacevkir: creating toDao() method for every object will lead to something elegant like this:

`nviCandidateRepository.save(pendingCandidate.toDao());`

truhacevkir (commented):
Some comments, in my opinion there is no need to call this classes for dto, since they do not contains additional fields, they are identical to our business objects.

truhacevkir (approved):
Ser bra ut, usikker om vi ønsker å ha dto's som er helt lik business objekter.
Inn i dao mappe ligger det mest sannsynligvis business objekter.

PR #29, opened 2023-08-17
Np 44967 renew json web token

truhacevkir:


template.yaml
```diff
@@ -224,9 +227,17 @@ Resources:
...
+      SnapStart:
+        ApplyOn: PublishedVersions
       Environment:
         Variables:
           EXPANDED_RESOURCES_BUCKET: !Ref ResourcesBucket
+      Events:
+        SqsEvent:
+          Type: SQS
+          Properties:
+            Queue: !GetAtt NewCandidateQueue.Arn
```
monaullah, line 240: Isn't UpsertNviCandidateHandler the consumer of this queue?

truhacevkir: Thats true, this one will listen to DynamoDb event, it is implemented in other pr

ketilaa (approved):

PR #31, opened 2023-08-21
Np 45135 dynamo db event handler

truhacevkir:
In my opinion StorageReader has to be moved to commons and we should have own test class for it.


index-handlers/src/main/java/no/sikt/nva/nvi/index/IndexNviCandidateHandler.java
```diff
@@ -20,14 +20,14 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class IndexNviCandidateHandler implements RequestHandler<SQSEvent, Void> {
+public class IndexNviCandidateHandler implements RequestHandler<DynamodbEvent, Void> {
```
truhacevkir, line 23: I think this should now be called UpdateIndexHandler or something like this.

brinxmat: Definitely change the name since you're checking if it is a Insert/Modify/Delete, but do this in a new PR that covers just this. Naming-wise, it is probably clearest to use `UpdateIndexDocumentHandler`

index-handlers/src/main/java/no/sikt/nva/nvi/index/utils/NviCandidateIndexDocumentGenerator.java
```diff
@@ -42,30 +41,30 @@ public final class NviCandidateIndexDocumentGenerator {
...
     }
 
-    public static NviCandidateIndexDocument generateNviCandidateIndexDocument(String resource,
-                                                                              NviCandidateMessageBody candidate) {
-        return createNviCandidateIndexDocument(
-            attempt(() -> dtoObjectMapper.readTree(resource)).map(root -> root.at("/body")).orElseThrow(),
-            candidate.affiliationApprovals());
+    public static NviCandidateIndexDocument generateDocument(String resource, List<String> affiliationApprovals) {
+        return createNviCandidateIndexDocument(attempt(() -> dtoObjectMapper.readTree(resource))
+                                                   .map(root -> root.at("/body")).orElseThrow(),
```
brinxmat, line 46: Extract constant for `/body`

brinxmat (changes requested):

template.yaml
```diff
@@ -236,6 +236,14 @@ Resources:
       Timeout: 30
       MemorySize: 1536
       Role: !GetAtt NvaNviRole.Arn
+      Events:
```
monaullah, line 239: I think there have been som copy/paste issues between UpsertNviCandidateHandler and IndexNviCandidateHandler in this template. The UpsertNviCandidateHandler properties seem wrong

monaullah: IndexNviCandidateHandler should be the one listening til the dynamoDB events

index-handlers/src/main/java/no/sikt/nva/nvi/index/model/NviCandidateMessageBody.java
```diff
@@ -6,7 +6,9 @@
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.net.URI;
 import java.util.List;
+import nva.commons.core.JacocoGenerated;
 
+@JacocoGenerated
 @JsonAutoDetect(fieldVisibility = Visibility.ANY)
 @JsonSerialize
 public record NviCandidateMessageBody(@JsonProperty(PUBLICATION_BUCKET_URI) String publicationBucketUri,
```
monaullah, line 14: I think this class can be deleted now?

truhacevkir: It cant, this object is never tested in commons

monaullah (changes requested):
See comment in template

index-handlers/src/main/java/no/sikt/nva/nvi/index/model/NviCandidateMessageBody.java
truhacevkir: It is still used in indexHandler, cause dynamoDbevent does not contain all the data we need to index

index-handlers/src/main/java/no/sikt/nva/nvi/index/aws/OpenSearchClient.java
```diff
@@ -80,14 +80,14 @@ public void addDocumentToIndex(NviCandidateIndexDocument indexDocument) {
     }
 
     @Override
-    public void removeDocumentFromIndex(NviCandidateIndexDocument indexDocument) throws IOException {
-        client.withTransportOptions(getOptions()).delete(contructDeleteRequest(indexDocument));
+    public void removeDocumentFromIndex(NviCandidateIndexDocument indexDocument) {
+        attempt(() -> client.delete(contructDeleteRequest(indexDocument))).orElseThrow();
```
monaullah, line 84: Why have you removed `client.withTransportOptions(getOptions())`? Don't we need it for renewing jwt?

index-handlers/src/main/java/no/sikt/nva/nvi/index/utils/NviCandidateIndexDocumentGenerator.java
```diff
@@ -124,6 +120,17 @@ private static String extractYear(JsonNode resource) {
         return extractJsonNodeTextValue(resource, JSON_PTR_PUBLICATION_DATE_YEAR);
     }
 
+    private static String extractJsonNodeTextValue(JsonNode node, String jsonPointer) {
```
monaullah, line 123: This method is moved to JsonUtils in commons

monaullah (approved):

PR #38, opened 2023-08-24
Np 45147 calculate points academic chapter

monaullah:
Part 2 for jira issue: https://unit.atlassian.net/browse/NP-45105

For AcademicChapter, publication channel type is relevant for the nvi point calculation. Added channelType as a factor in the point calculation. The INSTANCE_TYPE_AND_LEVEL_POINT_MAP is quite complex, I am open for suggestions to calculate points in a smoother way.

PS: Still keeping variable names long to make it clear what they represent.


truhacevkir:
We may add some test data from cristin to verify correct calculation of points

monaullah:
> 

I agree, going through Cristin calculations and adding more tests atm, can make new PR with more Cristin-examples @truhacevkir

hackon (approved):

PR #39, opened 2023-08-25
Np 45128 update period endpoint

truhacevkir:


docs/openapi.yaml
```diff
@@ -85,6 +85,57 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Problem"
+    put:
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/NviPeriod'
```
anetteOlli, line 94: Consider using the request body evaluators:
x-amazon-apigateway-request-validators

anetteOlli: This gives you the perk of faster bad request responses. -> bad request response ~ 5ms
And less need for validating the request body inside the lambda. -> faster 201 response
And this ensures that your openapi spec is not just theoretical, it's part of the actual implementation.

anetteOlli (changes requested):

docs/openapi.yaml
anetteOlli: aws request validation documentation: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-request-validation-sample-api-swagger.html

truhacevkir: Have added body required, but it seems that there is no string validation, only number value can be validated.

nvi-rest/src/main/java/no/sikt/nva/nvi/rest/UpdateNviPeriodHandler.java
```diff
@@ -0,0 +1,47 @@
...
+import nva.commons.apigateway.ApiGatewayHandler;
+import nva.commons.apigateway.RequestInfo;
+import nva.commons.apigateway.exceptions.ApiGatewayException;
+
+
+public class UpdateNviPeriodHandler extends ApiGatewayHandler<NviPeriodDto, NviPeriodDto> {
+
+    private final NviService nviService;
+
+    public UpdateNviPeriodHandler(NviService nviService) {
```
torarnet, line 23: Does a handler need a no param constructor?

torarnet (approved):
See comment

nvi-rest/src/main/java/no/sikt/nva/nvi/rest/UpdateNviPeriodHandler.java
truhacevkir: It does, but for now we do not have defaultService to use. I add default constructor with null service

PR #48, opened 2023-08-29
Persist candidate points

monaullah:
Added `points` field in `Candidate`. Persists points with Candidate in NviService

### DIFF
--- a/nvi-test/src/main/java/no/sikt/nva/nvi/test/TestUtils.java
+++ b/nvi-test/src/main/java/no/sikt/nva/nvi/test/TestUtils.java
@@ -1,5 +1,11 @@
 package no.sikt.nva.nvi.test;
 
+import static no.unit.nva.testutils.RandomDataGenerator.randomBoolean;
+import static no.unit.nva.testutils.RandomDataGenerator.randomElement;
+import static no.unit.nva.testutils.RandomDataGenerator.randomInteger;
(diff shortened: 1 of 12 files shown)


nvi-test/src/main/java/no/sikt/nva/nvi/test/TestUtils.java
```diff
@@ -106,7 +92,8 @@ public static Candidate randomCandidate() {
     }
 
     public static ApprovalStatus randomApprovalStatus() {
-        return new ApprovalStatus(randomUri(), randomElement(Status.values()), randomUsername(), Instant.EPOCH);
+        return new ApprovalStatus(randomUri(), randomElement(Status.values()), BigDecimal.ONE, randomUsername(),
```
sondrev, line 95: Should use bigger values for BigDecimal - on both side of the decimal . That way we can test that the conversion back/forth dynamodb works as expected in CandidateTest.
BigDecimal<->String<->AttributeValue(number)

monaullah: Added `randomBigDecimal()` for testing with a random values and a scale of 10. The specs for nvi points requires a scale of 4, but might be changed to 7 in the future.

sondrev (approved):

sondrev (approved):

PR #61, opened 2023-09-01
Candidate upsert update

sondrev:


nvi-commons/src/test/java/no/sikt/nva/nvi/common/service/NviServiceTest.java
```diff
@@ -180,16 +208,16 @@ private static Username randomUsername() {
         return new Username(randomString());
     }
 
-    private CandidateEvaluatedMessage createEvaluatedCandidateDto(UUID identifier,
+    private CandidateEvaluatedMessage createEvaluatedCandidateDto(UUID bucketIdentifier,
```
truhacevkir, line 211: Maybe rename to bucketUri

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/NviService.java
```diff
@@ -47,8 +47,11 @@ public static NviService defaultNviService() {
...
     @JacocoGenerated
     public Optional<CandidateWithIdentifier> upsertCandidate(CandidateEvaluatedMessage evaluatedCandidate) {
-        if (isNotExistingCandidate(evaluatedCandidate) && isNviCandidate(evaluatedCandidate)) {
+        if (!isExistingCandidate(evaluatedCandidate) && isNviCandidate(evaluatedCandidate)) {
             return Optional.of(createCandidate(evaluatedCandidate));
+        } else if (isExistingCandidate(evaluatedCandidate) && isNviCandidate(evaluatedCandidate)) {
+            var existing = findByPublicationId(evaluatedCandidate.candidateDetails().publicationId()).orElseThrow();
+            return Optional.of(updateCandidate(existing.identifier(), evaluatedCandidate));
         }
         return Optional.empty();
```
truhacevkir, line 56: Can be written in more attempt() way, but it is up to you :)
```
return doesNotExist(evaluatedCandidate)
           ? Optional.of(createCandidate(evaluatedCandidate))
           : Optional.of(evaluatedCandidate.candidateDetails().publicationId())
                 .map(this::findByPublicationId)
                 .map(Optional::orElseThrow)
                 .map(CandidateWithIdentifier::identifier)
                 .map(id -> updateCandidate(id, evaluatedCandidate))
                 .or(Optional::empty);
```

truhacevkir (approved):

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/NviService.java
```diff
@@ -47,8 +47,11 @@ public static NviService defaultNviService() {
     //TODO: Remove JacocoGenerated when other if/else cases are implemented
     @JacocoGenerated
     public Optional<CandidateWithIdentifier> upsertCandidate(CandidateEvaluatedMessage evaluatedCandidate) {
-        if (isNotExistingCandidate(evaluatedCandidate) && isNviCandidate(evaluatedCandidate)) {
+        if (!isExistingCandidate(evaluatedCandidate) && isNviCandidate(evaluatedCandidate)) {
             return Optional.of(createCandidate(evaluatedCandidate));
+        } else if (isExistingCandidate(evaluatedCandidate) && isNviCandidate(evaluatedCandidate)) {
+            var existing = findByPublicationId(evaluatedCandidate.candidateDetails().publicationId()).orElseThrow();
```
monaullah, line 53: There are some more business rules if we want to Update an existing Candidate. See: [Reset NVI candidate on change](https://unit.atlassian.net/browse/NP-45113).

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/NviService.java
sondrev: Thanks for the input and the code example. It would be better if the code was finished, but now we are missing some cases, and I think it will be more efficient to refactor it after we have accounted for all cases

PR #92, opened 2023-09-18
NP-45218-rejected-note

hackon:


nvi-rest/src/main/java/no/sikt/nva/nvi/rest/upsert/UpdateNviCandidateStatusHandler.java
```diff
@@ -57,6 +59,12 @@ protected Integer getSuccessStatusCode(NviStatusRequest input, CandidateResponse
         return HTTP_OK;
     }
 
+    private void verifyHasReasonWhenRejecting(NviStatusRequest input) throws BadRequestException {
+        if (input.status() == NviApprovalStatus.REJECTED && input.reason() == null) {
+            throw new BadRequestException("Missing reason for rejection");
+        }
```
truhacevkir, line 65: Shouldn't it be:
`input.reason() != null` 
Use `nonNull(input.reason())` instead of `!=`

nvi-rest/src/main/java/no/sikt/nva/nvi/rest/upsert/UpdateNviCandidateStatusHandler.java
```diff
@@ -65,7 +73,7 @@ private void verifyRequesteeIsCorrectCustomer(URI institutionId, URI currentCust
 
     private DbApprovalStatus toStatus(NviStatusRequest input, DbUsername username) {
         return new DbApprovalStatus(input.institutionId(), mapStatus(input.status()),
-                                    username, username, Instant.now());
+                                    username, username, Instant.now(), input.reason());
     }
```
truhacevkir, line 77: What about using builder instead of constructor. I mean we already have builder for this object

nvi-rest/src/test/java/no/sikt/nva/nvi/rest/upsert/UpdateNviCandidateStatusHandlerDbTest.java
```diff
@@ -63,14 +68,74 @@ void shouldUpdateApprovalStatus(NviApprovalStatus status) throws IOException {
...
+        var institutionId = randomUri();
+        var dbCandidate = createCandidate(institutionId);
+        var candidate = nviCandidateRepository.create(dbCandidate,
+                                                      List.of(DbApprovalStatus.builder()
+                                                                  .institutionId(institutionId)
+                                                                  .status(DbStatus.REJECTED)
+                                                                  .reason("I made a mistake")
+                                                                  .build()));
+
+        var req = new NviStatusRequest(candidate.identifier(),
```
truhacevkir, line 127: What about naming `req` variables as requestBody, think it as a little bit confusing to have two parameters which have almost the same names next to each other.

PR #108, opened 2023-09-25
Np 45218 Rejection note when rejecting candidate

monaullah:
Added `reason` to ApprovalStatus, which is required if status is updated to rejected


nvi-commons/src/main/java/no/sikt/nva/nvi/common/db/model/DbApprovalStatus.java
```diff
@@ -78,11 +84,16 @@ private DbApprovalStatus resetStatus(NviService nviService) {
     private DbApprovalStatus finalizeStatus(NviService nviService, UpdateStatusRequest request) {
         var approval = this.fetch(nviService);
         var username = DbUsername.fromString(request.username());
+        boolean isRejectionRequest = isRejection(request);
+        if (isRejectionRequest && isNull(request.reason())) {
+            throw new UnsupportedOperationException("Cannot reject approval status without reason.");
+        }
```
truhacevkir, line 90: I would not instantiate boolean parameter, but using isRejection(request) function directly. Maybe extract error message to constant.

truhacevkir (commented):
Ser bra ut.

sondrev (approved):

PR #141, opened 2023-10-10
Curator search has default value if not set, and if searched for outs…

sondrev:
…ide of allowed scope will return error

### DIFF
--- a/index-handlers/src/main/java/no/sikt/nva/nvi/index/SearchNviCandidatesHandler.java
+++ b/index-handlers/src/main/java/no/sikt/nva/nvi/index/SearchNviCandidatesHandler.java
@@ -8,13 +8,17 @@ import static no.sikt.nva.nvi.index.utils.PaginatedResultConverter.toPaginatedRe
 import static no.unit.nva.commons.json.JsonUtils.dtoObjectMapper;
 import static nva.commons.core.attempt.Try.attempt;
 import com.amazonaws.services.lambda.runtime.Context;
+import com.google.common.collect.Sets;
 import java.net.HttpURLConnection;
 import java.net.URI;
 import java.time.ZonedDateTime;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import no.sikt.nva.nvi.index.aws.SearchClient;
 import no.sikt.nva.nvi.index.model.CandidateSearchParameters;
 import no.sikt.nva.nvi.index.model.NviCandidateIndexDocument;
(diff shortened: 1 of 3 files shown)


nvi-commons/src/main/java/no/sikt/nva/nvi/common/utils/GraphUtils.java
```diff
@@ -19,6 +19,9 @@
 import org.slf4j.LoggerFactory;
 
 public class GraphUtils {
+    public static final String APPLICATION_JSON = "application/json";
+    public static final String PART_OF_PROPERTY = "https://nva.sikt.no/ontology/publication#partOf";
+    public static final String HAS_PART_PROPERTY = "https://bibsysdev.github.io/src/organization-ontology.ttl#hasPart";
```
sondrev, line 24: @brinxmat We need to fix this

sondrev: We should change the datamodel so we can use https://nva.sikt.no/ontology/publication#hasPart

index-handlers/src/main/java/no/sikt/nva/nvi/index/SearchNviCandidatesHandler.java
```diff
@@ -49,20 +68,45 @@ protected PaginatedSearchResult<NviCandidateIndexDocument> processInput(Void inp
...
         var username = requestInfo.getUserName();
         var year = extractQueryParamPublicationDateOrDefault(requestInfo);
 
+        assertUserIsAllowedToSearchAffiliations(affiliations, topLevelOrg);
+
         var candidateSearchParameters = new CandidateSearchParameters(affiliations, excludeSubUnits, filter, username,
-                                                                      year, customer, offset, size);
+                                                                      year, topLevelOrg, offset, size);
 
         return attempt(() -> openSearchClient.search(candidateSearchParameters))
```
truhacevkir, line 83: I think we will soon need a method that extracts all searchparams from requestInfo, something like:

`CandidateSearchParams.fromRequestInfo(requestInfo)`

index-handlers/src/test/java/no/sikt/nva/nvi/index/OpenSearchClientTest.java
```diff
@@ -384,10 +391,12 @@ private static PublicationDetails randomPublicationDetails() {
                                       List.of());
     }
 
-    private static PublicationDetails randomPublicationDetailsWithYear(String year) {
+    private static PublicationDetails randomPublicationDetailsWithYearAndCurator(String year, String affiliation) {
```
monaullah, line 394: `randomPublicationDetailsWithYearAndContributor`

index-handlers/src/test/java/no/sikt/nva/nvi/index/OpenSearchClientTest.java
```diff
@@ -138,11 +143,11 @@ void shouldReturnDocumentsFromIndexAccordingToGivenOffsetAndSize() throws IOExce
 
     @Test
     void shouldReturnEverythingIfYearFilterAndAffiliationsAreNotSet() throws IOException, InterruptedException {
```
monaullah, line 145: Confused by the test name and searchParameters in the test. The test is called `shouldReturnEverythingIfYearFilterAndAffiliationsAreNotSet`, but you are adding affiliations as search params in the test?
Rename test?

index-handlers/src/test/java/no/sikt/nva/nvi/index/SearchNviCandidatesHandlerTest.java
```diff
@@ -3,10 +3,11 @@
 import static no.sikt.nva.nvi.index.SearchNviCandidatesHandler.QUERY_PARAM_EXCLUDE_SUB_UNITS;
```
monaullah, line 3: Add another test `shouldReturnPaginatedSearchResultWithUserTopLevelOrgAsDefaultAffiliationIfNotSet()`?

PR #152, opened 2023-10-16
Refactor handling of non candidates

monaullah:
Added UpsertNonCandidateRequest for non candidates. Updated tests

PR #153, opened 2023-10-16
Np 45440 Undistributed points

monaullah:
Jira task: [Undistributed points - update index and GET candidate endpoint](https://unit.atlassian.net/browse/NP-45440)

- Added `undistributedPoints` in CandidateDto
- Removed `isApplicable`-check from `toDto` method in CandidateBO, does not belong there
- Modified NviCandidateIndexDocumentGenerator to update points in index with totalPoints stored in DB, instead of adding sum of all institution points


nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/CandidateBO.java
```diff
@@ -436,6 +443,14 @@ private static String mapToUsernameString(Username assignee) {
...
     }
 
+    private BigDecimal calculateUndistributedPoints() {
+        return totalPoints.subtract(sumDistributedPoints()).setScale(POINTS_SCALE, ROUNDING_MODE);
+    }
+
+    private BigDecimal sumDistributedPoints() {
+        return institutionPoints.values().stream().reduce(BigDecimal.ZERO, BigDecimal::add);
+    }
+
```
truhacevkir, line 453: I think fronted need both points and undistributedPoints. Maybe rename points to distrubtedPoints

truhacevkir (approved):

brinxmat (commented):
Did have a comment about checking instance of Candidate/Non-candidates, but it seems to have disappeared in an interim recfactor.

monaullah:
> Did have a comment about checking instance of Candidate/Non-candidates, but it seems to have disappeared in an interim recfactor.

That was another PR: https://github.com/BIBSYSDEV/nva-nvi/pull/152#discussion_r1361744327

PR #164, opened 2023-10-25
fix(nvi-commons): fix failing write throttle issue in batch scan #NP-45508

torbjokv:


nvi-commons/src/main/java/no/sikt/nva/nvi/common/db/DynamoDbRetryWrapper.java
```diff
@@ -0,0 +1,119 @@
...
+    private long calculateBackoffTime(int requestCount) {
+        return initialRetryWaitTimeMs * (long) Math.pow(2, requestCount - 1d);
+    }
+
+    private static void sleep(long waitTime) {
+        try {
+            Thread.sleep(waitTime);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
```
torbjokv, line 64: Maybe?
```suggestion
        try {
            Thread.sleep(waitTime);
        } catch (InterruptedException e) {
            Thread.currentThread().interrupt();
        }
```

monaullah (approved):

PR #165, opened 2023-10-25
Adding contributor and assignee filter params in NVI candidates search.

EyobSikt:
Adding contributor and assignee filter params in NVI candidates search

### DIFF
--- a/index-handlers/src/main/java/no/sikt/nva/nvi/index/SearchNviCandidatesHandler.java
+++ b/index-handlers/src/main/java/no/sikt/nva/nvi/index/SearchNviCandidatesHandler.java
@@ -45,6 +45,8 @@ public class SearchNviCandidatesHandler
     private static final int DEFAULT_OFFSET_SIZE = 0;
     public static final String QUERY_PARAM_YEAR = "year";
     public static final String QUERY_PARAM_CATEGORY = "category";
+    public static final String QUERY_PARAM_CONTRIBUTOR = "contributor";
+    public static final String QUERY_PARAM_ASSIGNEE = "assignee";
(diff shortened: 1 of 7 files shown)


index-handlers/src/main/java/no/sikt/nva/nvi/index/SearchNviCandidatesHandler.java
```diff
@@ -171,6 +176,16 @@ private static String extractQueryParamTitle(RequestInfo requestInfo) {
             .getOrDefault(QUERY_PARAM_TITLE, null);
     }
 
+    private static String extractQueryParamContributor(RequestInfo requestInfo) {
```
monaullah, line 179: If these parameters don't have a default value, there is no need to use `.getOrDefault()`, you can just use `get()`:
`return requestInfo.getQueryParameters(QUERY_PARAM_CONTRIBUTOR).get()`

index-handlers/src/main/java/no/sikt/nva/nvi/index/utils/PaginatedResultConverter.java
```diff
@@ -86,6 +91,12 @@ private static Map<String, String> getQueryParameters(List<URI> affiliations, bo
         if (isNotNullTitle(title)) {
             queryParams.put(QUERY_PARAM_TITLE, title);
         }
+        if (isNotNullContributor(contributor)) {
```
monaullah, line 94: Replace with `Objects.nonNull` method?
```
if (nonNull(contributor)) {
    queryParams.put(QUERY_PARAM_CONTRIBUTOR, contributor);
}
if (nonNull(assignee)) {
    queryParams.put(QUERY_PARAM_ASSIGNEE, assignee);
}
```

monaullah (approved):

PR #168, opened 2023-11-01
Np 45466 fetch candidate by publication

EyobSikt:
Fetch candidate by publicationId

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -229,7 +229,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Problem"
-  /candidate/{candidateIdentifier}:
+  /candidate/by-identifier/{candidateIdentifier}:
     get:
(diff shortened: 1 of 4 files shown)


docs/openapi.yaml
```diff
@@ -270,6 +270,47 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Problem"
+  /candidate/publication/{candidatePublicationId+}:
+    get:
+      description: 'Get scientific index candidate by publication identifier'
+      parameters:
+        - in: path
+          name: candidatePublicationId+
+          description: The UTF-8 Encoded pubication identifer of the candidate
```
monaullah, line 279: Is the type of path parameter candidatePublicationId a URI or an UUID identifier? If it is a URI, make it clearer? (You are testing with a URI in FetchNviCandidateByPublicationIdHandlerTest)

monaullah (approved):
Nice!

PR #196, opened 2023-12-13
Np 45509 DataEntryUpdateHandler

monaullah:
First, simple version of `DataEntryUpdateHandler`. Only implemented handling of dynamoDb events for `CandidateDao`. See todos, handling of rest of the dao types coming in next PR.


torbjokv (approved):
As long as "private static AttributeValue getAttributeValue"stays only in tests and code is not active before todos are done, this looks good to me :-)

PR #200, opened 2023-12-19
Np 45509 IndexDocumentHandler: Send event after persisting index document

monaullah:
The plan was to listen to S3 events, but this is not possible when the bucket is in another stack. Therefore, added new functionality in IndexDocumentHandler:
- send event with s3 uri when document is persisted


ketilaa (approved):

torbjokv (approved):
Probably possible to listen to s3 via eventbridge, but works great this also :-) https://stackoverflow.com/questions/56503098/how-to-create-s3-and-triggered-lambda-in-2-different-cloudformation-templates Could maybe become a bit to much not necessary events events when we have this one big bucket for everything.

PR #215, opened 2024-01-04
Add json sub type for CandidateUniquenessEntryDao

monaullah:


torbjokv (approved):

sondrev (approved):
Code changes are fine, but consider if we need UniquenessEntries as Dao's, as they are not in publication-api where we copied the pattern from

monaullah:
I was a bit to quick with this PR, working on a test to verify bug and fix.

PR #217, opened 2024-01-05
Fix attribute value bug

monaullah:
Jira bug: [Candidate modifications/approvals are not getting indexed](https://unit.atlassian.net/browse/NP-45929)

Reason: `AttributeValue`s with null-values where not parsed correctly with `dynamoDbMapper` (ObjectMapper) in `DynamoDbUtils`, causing failures in `DynamoEntryWithRangeKey.parseAttributeValuesMap()`.

Not a fan of this solution, its very aws dynamoDb model-specific. Feel free to suggest better ways to handle this issue.

### DIFF
--- a/event-handlers/build.gradle
+++ b/event-handlers/build.gradle
@@ -36,6 +36,7 @@ dependencies {
     implementation libs.bundles.logging
     implementation libs.bundles.jena
     testImplementation project(':nvi-test')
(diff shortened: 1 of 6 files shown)


ketilaa (approved):
It is what it is. Dynamo on AWS.

brinxmat (approved):

PR #221, opened 2024-01-08
shouldNotBuildIndexDocumentIfCandidateIsNotApplicable

monaullah:
To avoid race condition between `IndexDocumentHandler` and `RemoveIndexDocumentHandler` in the case candidate -> non candidate (see diagram), added a check in `IndexDocumentHandler` if the candidate is applicable before producing index document.
Screenshot 2024-01-08 at 14 19 18

### DIFF
--- a/index-handlers/src/main/java/no/sikt/nva/nvi/index/IndexDocumentHandler.java
+++ b/index-handlers/src/main/java/no/sikt/nva/nvi/index/IndexDocumentHandler.java
@@ -122,6 +122,10 @@ public class IndexDocumentHandler implements RequestHandler<SQSEvent, Void> {
         LOGGER.error(ERROR_MESSAGE, getStackTrace(exception));
     }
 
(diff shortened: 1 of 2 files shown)


index-handlers/src/main/java/no/sikt/nva/nvi/index/IndexDocumentHandler.java
```diff
@@ -178,6 +182,9 @@ private IndexDocumentWithConsumptionAttributes generateIndexDocument(DynamodbStr
     private IndexDocumentWithConsumptionAttributes generateIndexDocumentWithConsumptionAttributes(
         DynamodbStreamRecord record) {
         var candidate = fetchCandidate(record);
+        if (isNotApplicable(candidate)) {
+            return null;
+        }
```
brinxmat, line 187: Might have been tempted to check isApplicable or return null. Probably…

brinxmat (approved):

PR #224, opened 2024-01-12
Fix: Nvi candidates with integer points being reset on update when nothing has changed

monaullah:
Jira bug: [Nvi candidates with integer points being reset on update when nothing has changed](https://unit.atlassian.net/browse/NP-45980)

New:
- Always persist BigDecimals with scale 4, rounding half up
- When comparing BigDecimals, first set same scale and rounding mode


nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/Candidate.java
```diff
@@ -311,7 +312,21 @@ private static boolean publicationYearIsUpdated(UpsertCandidateRequest request,
     }
 
     private static boolean pointsAreUpdated(UpsertCandidateRequest request, CandidateDao existingCandidateDao) {
-        return !Objects.equals(request.institutionPoints(), mapToPointsMap(existingCandidateDao));
+        return existingCandidateDao.candidate()
+                   .points()
+                   .stream()
+                   .anyMatch(institutionPoints -> !equalsIgnoringScaleAndRoundingMode(
```
brinxmat, line 318: Negate in the method

brinxmat (approved):

PR #234, opened 2024-01-22
Np 46040 nvi admin search

truhacevkir:
Adding support to be able to search for all nvi candidates when nvi-admin.

Main changes:

Removing "mandatory default" customer filter when searching nvi candidates without providing affilition/customer.
Removing "mandatory default" year filter when searching nvi candidates without providing year.

This makes it possible for nvi-admin list nvi-candidates and get same "view" as nvi-curators at other institutions, in addition to list all nvi candidates.

### DIFF
--- a/event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinMapper.java
+++ b/event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinMapper.java
@@ -12,9 +12,7 @@ import no.sikt.nva.nvi.common.db.CandidateDao.DbCandidate;
 import no.sikt.nva.nvi.common.db.CandidateDao.DbLevel;
 import no.sikt.nva.nvi.common.db.CandidateDao.DbPublicationDate;
 import no.sikt.nva.nvi.common.db.model.InstanceType;
-
 import nva.commons.core.Environment;
-import nva.commons.core.paths.UnixPath;
 import nva.commons.core.paths.UriWrapper;
 
 public final class CristinMapper {
--- a/event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinMapper.java
+++ b/event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinMapper.java
@@ -22,6 +22,7 @@ public final class CristinMapper {
     public static final String AFFILIATION_DELIMITER = ".";
     public static final String CRISTIN = "cristin";
     public static final String ORGANIZATION = "organization";
+    public static final String GZ_EXTENSION = ".gz";
 
     private CristinMapper() {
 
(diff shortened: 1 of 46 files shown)


index-handlers/src/main/java/no/sikt/nva/nvi/index/SearchNviCandidatesHandler.java
```diff
@@ -129,9 +143,12 @@ private void assertUserIsAllowedToSearchAffiliations(List<URI> affiliations, URI
...
 
-    private String extractQueryParamPublicationDateOrDefault(RequestInfo requestInfo) {
-        return requestInfo.getQueryParameters()
-            .getOrDefault(QUERY_PARAM_YEAR, String.valueOf(ZonedDateTime.now().getYear()));
+    private String extractQueryParamPublicationDateOrDefault(RequestInfo requestInfo, boolean isAdmin) {
+        var yearQueryParam =  requestInfo.getQueryParameters()
+            .getOrDefault(QUERY_PARAM_YEAR, null);
+        return isAdmin
+                   ? yearQueryParam
+                   : nonNull(yearQueryParam) ? yearQueryParam : String.valueOf(ZonedDateTime.now().getYear());
```
truhacevkir, line 151: admin can list all candidates not depending on year

event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinMapper.java
```diff
@@ -94,12 +93,16 @@ private static ZonedDateTime getZonedDateTime(CristinNviReport cristinNviReport)
...
     private static URI constructPublicationBucketUri(String publicationIdentifier) {
-        return UriWrapper.fromUri(PERSISTED_RESOURCES_BUCKET)
+        return UriWrapper.fromHost(PERSISTED_RESOURCES_BUCKET)
                    .addChild("resources")
-                   .addChild(publicationIdentifier)
+                   .addChild(publicationIdentifierKeyInS3Bucket(publicationIdentifier))
                    .getUri();
     }
 
+    private static String publicationIdentifierKeyInS3Bucket(String publicationIdentifier) {
```
sondrev, line 102: or just "withGzExtension"?

index-handlers/src/main/java/no/sikt/nva/nvi/index/SearchNviCandidatesHandler.java
sondrev: nested elvis (?:) are hard to read.
Will this do the same?

(isAdmin || nonNull(yearQueryParam))
  ? yearQueryParam
  : String.valueOf(ZonedDateTime.now().getYear())

sondrev: Also I feel like you are giving this method too much responibility compared with what the other "ExtractQueryParam**OrDefault" methods does. Could be better to just return the first value, and let the parent caller do the admin-check and decide which year to pass on

sondrev (approved):

PR #236, opened 2024-01-26
InstanceType from Enum to String

truhacevkir:
I think we want to keep `InstanceType` enum when validating candidacy, so InstanceType will still be used inn `nvi-eveluator`, but it will never be stored as enum in database.

### DIFF
--- a/nvi-commons/src/main/java/no/sikt/nva/nvi/common/db/CandidateDao.java
+++ b/nvi-commons/src/main/java/no/sikt/nva/nvi/common/db/CandidateDao.java
@@ -268,7 +268,7 @@ public final class CandidateDao extends Dao {
     public record DbCandidate(URI publicationId,
                               URI publicationBucketUri,
                               boolean applicable,
-                              InstanceType instanceType,
+                              String instanceType,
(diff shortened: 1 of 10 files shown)


nvi-commons/src/main/java/no/sikt/nva/nvi/common/db/CandidateDao.java
```diff
@@ -345,8 +345,17 @@ public Builder applicable(boolean applicable) {
...
+            public Builder instanceType(String instanceType) {
+                var enums = Arrays.stream(InstanceType.values()).toList();
+                var instanceTypeEnum = enums.stream()
+                                           .filter(value -> value.toString().equals(instanceType))
+                                           .findFirst();
+                if (instanceTypeEnum.isPresent()) {
+                    this.builderInstanceType = instanceTypeEnum.get().getValue();
+                } else {
+                    this.builderInstanceType = instanceType;
+                }
```
truhacevkir, line 358: This piece should be set back as it was after we have run batchScan in all environments.

monaullah (approved):

PR #237, opened 2024-01-29
Attempting to migrate instance type

truhacevkir:


nvi-commons/src/main/java/no/sikt/nva/nvi/common/db/CandidateDao.java
```diff
@@ -283,6 +283,18 @@ public record DbCandidate(URI publicationId,
                               BigDecimal totalPoints
     ) {
 
+        public String instanceType() {
```
monaullah, line 286: Mark with todo and deprecated

monaullah (approved):

PR #241, opened 2024-02-01
Fix type info bug

monaullah:
Fix bug introduced by: [Index document - add type info](https://github.com/BIBSYSDEV/nva-nvi/commit/2df96ccd14c6945bae8eaaa9893e0d7746f7ccf5)

- Change `PaginatedSearchResult` type back to `NviCandidateIndexDocument`
- Add type field with static value for `NviCandidateIndexDocument`

### DIFF
--- a/index-handlers/src/main/java/no/sikt/nva/nvi/index/model/NviCandidateIndexDocument.java
+++ b/index-handlers/src/main/java/no/sikt/nva/nvi/index/model/NviCandidateIndexDocument.java
@@ -1,11 +1,7 @@
 package no.sikt.nva.nvi.index.model;
 
-import static no.unit.nva.commons.json.JsonUtils.dtoObjectMapper;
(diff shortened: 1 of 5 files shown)


index-handlers/src/main/java/no/sikt/nva/nvi/index/model/NviCandidateIndexDocument.java
```diff
@@ -19,12 +15,9 @@
 
 @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
 @JsonSerialize
-@JsonTypeInfo(
-    use = JsonTypeInfo.Id.NAME,
-    property = "type")
-@JsonTypeName("NviCandidate")
```
truhacevkir, line 25: This annotation does not add `type` param to serialized object?

truhacevkir (approved):

PR #248, opened 2024-02-08
Adding year to imported candidate

truhacevkir:
Task: Store reportedYear/ reportedPeriod for NviCandidate. 

Problem:

Not possible to add `:PERIOD#somePeriod` to hashKey or rangeKey, cause they must be equal to be able to run dynamo.getItem(UniqueKey), otherwise we have to use dynamo.query() for fetching single item which is not a good idea.

Solution:

Today we store publicationDate.year as secondary index key in dynamo. This year will also be reportedYear/period since our periods are years, like 2023. This key will be set as publicationDate.year for nvi candidates created from NVA, but set to given value for imported nvi candidates from Cristin. 

Alternative solution is to create one more secondary index, but this is not a good idea, because given implementation we have now, this new secondary index will contain the same data as existing byYear index. 

For future "report candidates" functionality it will be enough to add for example `isReported` boolean or enum and iterate through all candidates for period/year and set isReported value.

### DIFF
--- a/event-handlers/src/test/java/no/sikt/nva/nvi/events/db/DataEntryUpdateHandlerTest.java
+++ b/event-handlers/src/test/java/no/sikt/nva/nvi/events/db/DataEntryUpdateHandlerTest.java
@@ -9,6 +9,7 @@ import static no.sikt.nva.nvi.test.TestUtils.randomApproval;
 import static no.sikt.nva.nvi.test.TestUtils.randomCandidate;
 import static no.sikt.nva.nvi.test.TestUtils.randomCandidateBuilder;
 import static no.sikt.nva.nvi.test.TestUtils.randomUsername;
+import static no.sikt.nva.nvi.test.TestUtils.randomYear;
 import static no.unit.nva.testutils.RandomDataGenerator.randomElement;
 import static no.unit.nva.testutils.RandomDataGenerator.randomInstant;
 import static no.unit.nva.testutils.RandomDataGenerator.randomString;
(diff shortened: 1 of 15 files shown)


event-handlers/src/test/java/no/sikt/nva/nvi/events/db/DataEntryUpdateHandlerTest.java
```diff
@@ -193,11 +193,12 @@ private static PublishRequest createExpectedPublishedMessage(String message, Str
     }
 
     private static CandidateDao randomCandidateDao() {
-        return new CandidateDao(UUID.randomUUID(), randomCandidate(), UUID.randomUUID().toString());
+        return new CandidateDao(UUID.randomUUID(), randomCandidate(), UUID.randomUUID().toString(), randomString());
```
monaullah, line 196: `Year.now().toString()`?

nvi-commons/src/main/java/no/sikt/nva/nvi/common/db/CandidateDao.java
```diff
@@ -38,23 +39,28 @@
...
     public static final String TYPE = "CANDIDATE";
+    public static final String PERIOD_FIELD = "period";
     @JsonProperty(IDENTIFIER_FIELD)
     private final UUID identifier;
     @JsonProperty(DATA_FIELD)
     private final DbCandidate candidate;
     @JsonProperty(VERSION_FIELD)
     private final String version;
+    @JsonProperty(PERIOD_FIELD)
+    private final String period;
```
monaullah, line 50: Is the data type a String because it may be changes to some identifier instead of year in the future?

monaullah (approved):

PR #249, opened 2024-02-08
Migrate candidates to include createdDate and modifiedDate

monaullah:


truhacevkir (approved):

nvi-commons/src/main/java/no/sikt/nva/nvi/common/db/CandidateDao.java
```diff
@@ -304,6 +305,26 @@ public String instanceType() {
             }
         }
 
+        @Deprecated
+        //TODO: Remove after migration
```
brinxmat, line 309: Just add 

```
    @Deprecated
    private static Instant migrationMethodName(Instant instant) {
        if (isNull(instant)) {
            return Instant.now();
        } else {
            return instant;
        }
    }
```
Use the method in the getters and builders. The method can be edited afterwards to return the input and inlined.

brinxmat (changes requested):

PR #257, opened 2024-02-15
throwing exception when missing DbLevel

truhacevkir:
Assignees kan mangle i følge Terje


event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinMapper.java
```diff
@@ -92,7 +93,7 @@ private static DbLevel extractLevel(CristinNviReport cristinNviReport) {
                    .map(list -> list.get(0))
                    .map(ScientificResource::getQualityCode)
                    .map(DbLevel::parse)
-                   .orElse(null);
+                   .orElseThrow();
```
monaullah, line 96: Test that verifies that excepted exception is thrown?

monaullah (changes requested):

PR #271, opened 2024-03-20
NP-46476 Index document, global approval status Dispute

monaullah:
Jira issue: [Add Dispute (tvist) as enum value for global approval status](https://unit.atlassian.net/browse/NP-46476)

Global approval status value `Dispute` (også kjent som tvist) is required in data export and institution reports. See `RAPPORTSTATUS` in [instruks](https://www.cristin.no/nvi-rapportering/nvi-veiledninger/hvordan_lese_nvi-kontrolldata.html).

- If at least two institutions disagree on whether the candidate should be approved or rejected (one approved approval and one rejected approval), set `globalApprovalStatus` `Dispute` in index document.

### DIFF
--- a/nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/GlobalApprovalStatus.java
+++ b/nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/GlobalApprovalStatus.java
@@ -1,0 +1,19 @@
+package no.sikt.nva.nvi.common.service.model;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+public enum GlobalApprovalStatus {
+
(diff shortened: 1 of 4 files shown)


nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/GlobalApprovalStatus.java
```diff
@@ -0,0 +1,19 @@
+package no.sikt.nva.nvi.common.service.model;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+public enum GlobalApprovalStatus {
+
+    APPROVED("Approved"), PENDING("Pending"), REJECTED("Rejected"), DISPUTE("Dispute");
```
brinxmat, line 7: I suspect that it is "In Dispute" or to keep the pattern "Disputed", but it is likely not important

brinxmat (approved):

PR #281, opened 2024-04-05
Fix: periodYear migration

monaullah:
Move migration code to `CandidateDao.Builder`. This is used when reading from dynamoDb


sondrev (approved):

monaullah:
> Should we also update candidate() and identifier() method in CandidateDao to getters?

We should consider it, yes. In the example in [docs](https://aws.amazon.com/blogs/developer/introducing-immutable-class-mapping-for-the-enhanced-dynamodb-client-in-aws-sdk-for-java-2-x/), the getters in the class are not prefixed with "get", I think this is the reason they are implemented like this.

PR #288, opened 2024-04-10
Np 46529 fix creator affiliation points batch scan

monaullah:
To avoid loosing information about `NviCandidate` `CreatorAffiliationPoints` in case of a RequeueDLQ situation. Added test `shouldRequeueCandidateWithoutLossOfInformation`


event-handlers/src/main/java/no/sikt/nva/nvi/events/model/NviCandidate.java
```diff
@@ -76,16 +75,6 @@ public PublicationDetails.PublicationDate publicationDate() {
         return mapToPublicationDate(date);
     }
 
-    private static List<InstitutionPoints> mapToInstitutionPoints(Map<URI, BigDecimal> institutionPoints) {
-        return institutionPoints.entrySet().stream()
-                   .map(entry -> new InstitutionPoints(entry.getKey(),
-                                                       entry.getValue(),
-                                                       //TODO: Implement when Candidate contains
```
monaullah, line 83: Basically, this is implemented in this PR

truhacevkir (approved):

PR #287, opened 2024-04-10
Np 46529 UpsertNviCandidateHandler: Persist points for creator affiliation

monaullah:
Jira task: NP-46529, Refactor points for affiliation calculation, persist in db.

Part 1 -> [Merged](https://github.com/BIBSYSDEV/nva-nvi/pull/279)
Part 2 -> [Merged](https://github.com/BIBSYSDEV/nva-nvi/pull/283)
This is part 3: Persist points for affiliation in `UpsertNviCandidateHandler`

- Change in `EvalutatorService`: instead of sending affiliation points as part of `NviCreator`, change data type for `institutionPoints` from `Map<URI, BigDecimal>` to `List<InstitutionPoints>` (containing `CreatorAffiliationPoints`)
- Changed interface `UpsertCandidateRequest`: change data type for `institutionPoints` from `Map<URI, BigDecimal>` to `List<InstitutionPoints>` (containing `CreatorAffiliationPoints`)
- `Candidate` (BO): Persist creator affiliation points when persisting candidate in DB
- `CandidateDao`: Added `DbCreatorAffiliationPoints` in `DbInstitutionPoints`

This change will require a new point calculation for all candidates in all environments. I can use `ReEvaluateNviCandidatesHandler` for this purpose.

Also, tested in sandbox. See CANDIDATE#54a8ff9f-4a22-41d9-9c1d-f47031f1eef4 in DB.

(Sorry about the size, a lot of tests where affected)

PR #279, opened 2024-04-05
Np 46529 PointCalculator, calculate points for affiliation

monaullah:
Jira task: NP-46529, Refactor points for affiliation calculation, persist in db.

This is part 1 -> calculate points for affiliation (combination of `NviCreator` and `NviAffiliation`)  in `PointCalculator`

New:
- Changed data type for `institutionPoints` in `PointCalculation` from `Map<URI, BigDecimal>` to `List<InstitutionPoints>`
- New class `InstitutionPoints` contains both points for institution (top level) and points for affiliation
- Calculate `AffiliationPoints` in `PointCalculator`

Refactor:
- Split `PointCalculator` into `PointService` and `PointCalculator`

PR #289, opened 2024-04-11
Np 46529 Index document, Points for creator affiliation

monaullah:
Jira task: NP-46529, Refactor points for affiliation calculation, persist in db.

Part 1 -> https://github.com/BIBSYSDEV/nva-nvi/pull/279
Part 2 -> https://github.com/BIBSYSDEV/nva-nvi/pull/283
Part 3 -> https://github.com/BIBSYSDEV/nva-nvi/pull/287

This is the last part: add creator affiliation points in index document, so it can be used for export in data-report-api

### DIFF
--- a/index-handlers/src/main/java/no/sikt/nva/nvi/index/model/InstitutionPoints.java
+++ b/index-handlers/src/main/java/no/sikt/nva/nvi/index/model/InstitutionPoints.java
@@ -1,6 +1,5 @@
 package no.sikt.nva.nvi.index.model;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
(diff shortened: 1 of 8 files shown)


index-handlers/src/main/java/no/sikt/nva/nvi/index/model/InstitutionPoints.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.List;
+
+@JsonSerialize
+@JsonTypeInfo(
+    use = JsonTypeInfo.Id.NAME,
+    property = "type")
+@JsonTypeName("InstitutionPoints")
+public record InstitutionPoints(URI institutionId,
+                                BigDecimal institutionPoints,
+                                @JsonProperty("breakdown") List<CreatorAffiliationPoints> creatorAffiliationPoints) {
```
truhacevkir, line 18: why not to use default jsonName?
breakdown?

truhacevkir (approved):

index-handlers/src/main/java/no/sikt/nva/nvi/index/model/InstitutionPoints.java
monaullah: I used breakdown when planning this task with @brinxmat, but probably just as good as `creatorAffiliationPoints`. @brinxmat What do you think?

brinxmat: The naming confused Kir, so choose a better one… creatorAffiliationPoints is likely good, yes.

brinxmat (approved):

PR #298, opened 2024-04-18
Np 46528 Aggregations, new type and refactoring

monaullah:
Jira issue: [NP-46528 Aggregations for approvals](https://unit.atlassian.net/browse/NP-46528)

This is part 2:
- Added new SearchAggregation: `ORGANIZATION_APPROVAL_STATUS_AGGREGATION` (this is not implemented yet, see TODO)
- Refactored `Aggregations.generateAggregations()` method to either generate all aggregations or just a specific aggregation.

### DIFF
--- a/index-handlers/src/main/java/no/sikt/nva/nvi/index/Aggregations.java
+++ b/index-handlers/src/main/java/no/sikt/nva/nvi/index/Aggregations.java
@@ -36,13 +36,9 @@ public final class Aggregations {
 
     public static Map<String, Aggregation> generateAggregations(String aggregationType, String username,
                                                                 String topLevelCristinOrg) {
-        var aggregations = new HashMap<String, Aggregation>();
-        if (aggregationTypeIsNotSpecified(aggregationType)) {
-            generateAllAggregationTypes(username, topLevelCristinOrg, aggregations);
-        } else {
-            generateSingleAggregation(aggregationType, username, topLevelCristinOrg, aggregations);
-        }
-        return aggregations;
+        return aggregationTypeIsNotSpecified(aggregationType)
+                   ? generateAllAggregationTypes(username, topLevelCristinOrg)
+                   : generateSingleAggregation(aggregationType, username, topLevelCristinOrg);
     }
 
     public static Query containsPendingStatusQuery() {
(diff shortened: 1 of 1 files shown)


index-handlers/src/main/java/no/sikt/nva/nvi/index/Aggregations.java
```diff
@@ -39,26 +28,20 @@
...
-                         finalizedCollaborationAggregation(customer, REJECTED));
-        aggregations.put(ASSIGNMENTS_AGG.getAggregationName(), assignmentsAggregation(username, customer));
-        aggregations.put(COMPLETED_AGGREGATION_AGG.getAggregationName(), completedAggregation(customer));
-        aggregations.put(TOTAL_COUNT_AGGREGATION_AGG.getAggregationName(), totalCountAggregation(customer));
+        if (aggregationTypeIsNotSpecified(aggregationType)) {
+            generateAllAggregationTypes(username, topLevelCristinOrg, aggregations);
+        } else {
+            generateSingleAggregation(aggregationType, username, topLevelCristinOrg, aggregations);
+        }
         return aggregations;
```
truhacevkir, line 45: It seems we don't need to initialize an empty map before the if-else conditional. Both generateAllAggregationTypes and generateSingleAggregation methods can directly return a new map.

index-handlers/src/main/java/no/sikt/nva/nvi/index/Aggregations.java
```diff
@@ -39,26 +28,20 @@
...
     public static final int MULTIPLE = 2;
+    public static final String ALL_AGGREGATIONS = "all";
     private static final CharSequence JSON_PATH_DELIMITER = ".";
 
     private Aggregations() {
     }
 
-    public static Map<String, Aggregation> generateAggregations(String username, String customer) {
+    public static Map<String, Aggregation> generateAggregations(String aggregationType, String username,
+                                                                String topLevelCristinOrg) {
```
truhacevkir, line 38: Looks good, but could you add an alternative to not to create aggregations? as in search-api -> aggregation=none

monaullah: We can add this later if we see the need 👍

PR #315, opened 2024-05-08
Add more cristin transfers

truhacevkir:
When creator misses approval for his institution and transferred_institution csv does not contain mapping for creator institution, we simply create approval for creators actual affiliation, which in most cases will not be top level org. In this way we will be able to import almost all nvi candidates from Cristin


event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinConversionException.java
```diff
@@ -11,7 +11,7 @@ public CristinConversionException(String message) {
     }
 
     public static CristinConversionException fromFailure(Failure<?> failure) {
-        var message = failure.getException().getMessage();
+        var message = failure.getException();
         return new CristinConversionException(String.format("%s %s", EXCEPTION_MESSAGE, message));
```
truhacevkir, line 15: Updating error report to contain whole exception, we need to know where exactly in code exception has occurred.

brinxmat (approved):
Not sure what the concrete outcome of this will affect — I'd let @monaullah take a look before merging

PR #317, opened 2024-05-15
candidate without approvals

truhacevkir:
When Nvi Candidate from Cristin is missing approvals, we create Nvi Candidate in nva without approvals and without points, as approvals in Cristin contains points per institution

### DIFF
--- a/event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinMapper.java
+++ b/event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinMapper.java
@@ -333,7 +333,7 @@ public final class CristinMapper {
                                                          departmentTransfer.getToInstitutionIdentifier())))
                    .findAny()
                    .map(CristinDepartmentTransfer::getToInstitutionIdentifier)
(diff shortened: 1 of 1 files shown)


event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinMapper.java
```diff
@@ -233,12 +233,16 @@ private static BigDecimal extractBasePoints(CristinNviReport cristinNviReport) {
...
+        if (institutions.isEmpty()) {
+            return List.of();
+        } else {
+            return getCreators(cristinNviReport).stream()
+                       .filter(CristinMapper::hasInstitutionPoints)
+                       .collect(collectToInstitutionOfPoints(institutions))
+                       .stream()
+                       .map(CristinMapper::toDbInstitutionPoints)
+                       .collect(Collectors.toList());
+        }
```
truhacevkir, line 245: When institutions/approvals for nvi candidate are empty, we create candidate without points and approvals.

torbjokv (approved):

PR #325, opened 2024-06-06
Np 46897 Evaluator, filter inactive period

monaullah:
- (Finally) Removed hardcoded year filter from nvi sparql (used to check if candidate meets requirements)
- Added check in evaluator: `isPublishedBeforeOrSameAsLatestClosedPeriod`
- Renamed `CandidateCalculator` to `CreatorVerificationUtil`, and moved requirement check (sparql query) to `EvaluatorService`
- `fetchAll()` method from `NviPeriod` to new class `NviPeriodService`


event-handlers/src/main/java/no/sikt/nva/nvi/events/evaluator/EvaluatorService.java
```diff
@@ -129,6 +146,22 @@ private static PublicationDate mapToPublicationDate(JsonNode publicationDateNode
                    .orElse(new PublicationDate(null, null, year.textValue()));
     }
 
+    private boolean isPublishedBeforeOrSameAsLatestClosedPeriod(PublicationDate publicationDate) {
```
truhacevkir, line 149: This method checks only if publicationYear is before latestClosedPeriod, isn't it? 
Could be renamed to something like `isWithinClosedPeriod`, checks if publication date falls within a closed period.

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/NviPeriod.java
```diff
@@ -54,17 +53,20 @@ public static NviPeriod update(UpdatePeriodRequest request, PeriodRepository per
...
-    }
-
     public static NviPeriod fetch(String publishingYear, PeriodRepository periodRepository) {
         return periodRepository.findByPublishingYear(publishingYear)
                    .map(NviPeriod::fromDbObject)
                    .orElseThrow(() -> PeriodNotFoundException.withMessage(
                        String.format("Period for year %s does not exist!", publishingYear)));
     }
 
+    public static NviPeriod fromDbObject(DbNviPeriod dbNviPeriod) {
```
truhacevkir, line 63: Maybe `fromDbNviPeriod()`

brinxmat (approved):

PR #330, opened 2024-06-19
Np 47020 FetchInstitutionStatusAggregationHandler

monaullah:
Jira task: _NP-47020 Allow NVI-curator to see NVI status for their institution_
From jira description: _The NVI status page is currently based on the same search as is used for the candidates list: GET /scientific-index/candidate. When we solve NP-47019 (Ensure curator only see NVI candidates for their viewingScope)
 the user will no longer see NVI-candidates they don’t have access to, meaning that the status will lose data for a lot of departments for many users._
_To solve this we need a new endpoint that returns all the aggregations currently located in organizationApprovalStatuses for the NVI aggregations...._

Changes:
- Repurposing old handler `FetchReportHandler` (not in use), renamed to `FetchInstitutionStatusAggregationHandler`
- `FetchInstitutionStatusAggregationHandler` returns aggregation `organizationApprovalStatuses` for users top level org and requested path param `year`
- Some package restructuring in `index` module
- Removed unused Excel classes in `rest` module


index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/FetchInstitutionStatusAggregationHandlerTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @BeforeEach
+    public void setUp() {
+        output = new ByteArrayOutputStream();
+        openSearchClient = mock(OpenSearchClient.class);
+        handler = new FetchInstitutionStatusAggregationHandler(openSearchClient);
+    }
+
+    @Test
+    void shouldReturnUnauthorizedWhenUserDoesNotHaveSufficientAccessRight() throws IOException {
```
brinxmat, line 53: RequiredAccessRight?

index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/FetchInstitutionStatusAggregationHandlerTest.java
```diff
@@ -0,0 +1,126 @@
...
+        handler.handleRequest(request, output, CONTEXT);
+        var response = GatewayResponse.fromOutputStream(output, String.class);
+        assertThat(response.getStatusCode(), is(Matchers.equalTo(HttpURLConnection.HTTP_OK)));
+        assertEquals(expectedNestedAggregation, response.getBody());
+    }
+
+    @Test
+    void shouldReturnNotFoundIfInstitutionDoesntExist() {
+
+    }
```
brinxmat, line 87: Rename `shouldPassWhenTestBodyDoesNotExist`?

brinxmat (changes requested):

PR #346, opened 2024-08-21
Np 47379 Fix: handle multiple channel types in evaluation

monaullah:
Jira issue: 
`EvaluateNVICandidateHandler` is expecting `publicationContext` `type` as String. In some cases, where there is a mismatch between NVA and Kanalregisteret on the channel type, publicationContext is array (See example).

```
{
  "id": "https://api.dev.nva.aws.unit.no/publication-channels-v2/series/123/2024",
  "type": [
    "Series",
    "Journal"
  ],
  "identifier": "C8CCD71B-FD8B-47B4-B72A-905F6219D7D5",
  "name": "Some name",
  "scientificValue": "LevelOne",
  "year": "2024"
}
```

Long term fix: NVA and Kanalregisteret should agree on the type of the channel
Fix short term: Revert this removal: https://github.com/BIBSYSDEV/nva-nvi/pull/277


brinxmat:
Do we know the root cause of this issue? Do we assume something about types that are encoded by kanalregisteret?

brinxmat (approved):
It fixed the issue, but as you point out the long term fix is needed

monaullah:
> Do we know the root cause of this issue? Do we assume something about types that are encoded by kanalregisteret?

I think it has to do with channel type imported via migration. Following up long term fix with Kir and Terje 👍

PR #356, opened 2024-09-03
Np 47488 IndexDocument: add publication channel id, type and scientificValue

monaullah:
Change in index document model: add `document.publicationDetails.publicationChannel` with `id`, `type` and `scientificValue`

Also did some refactoring in `NviIndexDocumentGenerator`: instead of getting `JsonNode expandedResource` and `Candidate candidate` as input parameters in method `generateDocument()`, added these as class variables.

### DIFF
--- a/index-handlers/src/main/java/no/sikt/nva/nvi/index/model/document/PublicationChannel.java
+++ b/index-handlers/src/main/java/no/sikt/nva/nvi/index/model/document/PublicationChannel.java
@@ -1,0 +1,71 @@
+package no.sikt.nva.nvi.index.model.document;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.net.URI;
+import java.util.Arrays;
+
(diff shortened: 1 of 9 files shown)


brinxmat (approved):

index-handlers/src/main/java/no/sikt/nva/nvi/index/model/document/PublicationChannel.java
```diff
@@ -0,0 +1,73 @@
...
+                                 String type,
+                                 ScientificValue scientificValue) {
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public enum ScientificValue {
+        LEVEL_ONE("LevelOne"),
+        LEVEL_TWO("LevelTwo");
```
LarsV123, line 19: Should it be possible to represent "LevelZero" as well, or is that not valid?

monaullah: That would be an illegal state for a `NviCandidate`, since `LevelOne` is a minimum requirement to qualify for nvi candidacy

truhacevkir (approved):

PR #360, opened 2024-09-05
Np 47490 Index document: Add page information

monaullah:
Jira task: NP-47490 Expand index document with page information

Extracting optional fields `begin`, `end` and `pages` from `publicationInstance` in expanded resource

Test that verifies this is `IndexDocumentHandlerTest.shouldBuildIndexDocumentAndPersistInS3WhenReceivingSqsEvent()` with `expectedIndexDocument` containing publication page info from test data setup (`expandedResource`)

### DIFF
--- a/index-handlers/src/main/java/no/sikt/nva/nvi/index/model/document/PublicationDetails.java
+++ b/index-handlers/src/main/java/no/sikt/nva/nvi/index/model/document/PublicationDetails.java
@@ -2,18 +2,18 @@ package no.sikt.nva.nvi.index.model.document;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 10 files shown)


index-handlers/src/main/java/no/sikt/nva/nvi/index/model/document/PublicationDetails.java
```diff
@@ -2,18 +2,18 @@
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.util.List;
 
 @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
 @JsonSerialize
-public record PublicationDetails(@JsonProperty("id") String id,
```
monaullah, line 11: The JsonProperty-annotations are not necessary here since the field name is the same

brinxmat (approved):

PR #362, opened 2024-09-05
Fix: Index document - Optional channel data

monaullah:
Indexing failing for candidates imported via Cristin, some of these do not have `channelType` or `channelId`

Changes:
- Handle null values in `IndexDocumentHandler` and `FetchInstitutionReportHandler`


brinxmat (approved):

brinxmat (approved):
See codacy

PR #374, opened 2024-09-13
Np 47564 source exclude non nvi contributors

monaullah:
Jira task: Export to NVI-kontrolldata optimization: exclude non-nvi-contributors from search response

Change:
- Add list of excludeFields in search parameters, use this to set `SourceConfig` in search request
- In `InstitutionReportGenerator`: exclude `publicationDetails.contributors` in search request

### DIFF
--- a/index-handlers/src/test/java/no/sikt/nva/nvi/index/aws/OpenSearchClientTest.java
+++ b/index-handlers/src/test/java/no/sikt/nva/nvi/index/aws/OpenSearchClientTest.java
@@ -1,5 +1,6 @@
 package no.sikt.nva.nvi.index.aws;
 
+import static java.util.Objects.requireNonNull;
 import static no.sikt.nva.nvi.index.model.document.ApprovalStatus.NEW;
 import static no.sikt.nva.nvi.index.model.document.ApprovalStatus.PENDING;
 import static no.sikt.nva.nvi.index.model.document.ApprovalStatus.REJECTED;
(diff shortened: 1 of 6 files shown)


index-handlers/src/test/java/no/sikt/nva/nvi/index/aws/OpenSearchClientTest.java
```diff
@@ -575,7 +577,25 @@ void organizationAggregationShouldNotContainAggregationsForOtherTopLevelOrgs()
         assertThat(orgIds, not(containsInAnyOrder(NTNU_INSTITUTION_ID.toString())));
     }
 
-    @NotNull
+    @Test
+    void shouldExcludeFields() throws IOException, InterruptedException {
+        var document = documentWithContributors();
+        addDocumentsToIndex(document);
+        var searchParameters = defaultSearchParameters()
+                                   .withExcludeFields(List.of("publicationDetails.contributors"))
```
LarsV123, line 585: Is this behavior intended to be the default for all queries, or only for queries that explicitly exclude contributors? If it's the default, perhaps add `.withExcludeFields(List.of("publicationDetails.contributors"))` directly to the helper method `defaultSearchParameters`?

index-handlers/src/test/java/no/sikt/nva/nvi/index/aws/OpenSearchClientTest.java
```diff
@@ -575,7 +577,25 @@ void organizationAggregationShouldNotContainAggregationsForOtherTopLevelOrgs()
...
+                                   .build();
+        var searchResponse = openSearchClient.search(searchParameters);
+        var firstHit = searchResponse.hits().hits().get(0).source();
+        assertNull(firstHit.publicationDetails().contributors());
```
LarsV123, line 589: This produces a linter warning because `publicationDetails` could be null. Not actually a problem, but perhaps add an assertion or wrapper to suppress the warning? There are a few similar cases in the same file too:

```
- Method invocation 'createdDate' may produce 'NullPointerException' :184
- Method invocation 'createdDate' may produce 'NullPointerException' :185
- Method invocation 'approvals' may produce 'NullPointerException': 360
- Method invocation 'publicationDetails' may produce 'NullPointerException': 582
```

LarsV123 (approved):

index-handlers/src/test/java/no/sikt/nva/nvi/index/aws/OpenSearchClientTest.java
monaullah: It's not intended to be the default, `publicationDetails.contributors` is still relevant for nvi search api

brinxmat (approved):

PR #382, opened 2024-09-24
Np 47651 InstitutionReportGenerator: Parallelize fetch

monaullah:
Changes:
- Fetch total number of candidates for given institution and year
- Split into batches of 1000
- Fetch batches in parallel


index-handlers/src/main/java/no/sikt/nva/nvi/index/aws/OpenSearchClient.java
```diff
@@ -148,14 +147,6 @@ public void createIndex() {
             .orElseThrow(failure -> handleFailure(ERROR_MSG_CREATE_INDEX, failure.getException()));
     }
 
-    private static void logSearchRequest(CandidateSearchParameters params) {
```
monaullah, line 151: Not relevant, this was just very noisy in the logs

brinxmat (approved):

PR #385, opened 2024-10-08
Np 47797 Index document, contributorsPreview

monaullah:
Jira task - Search optimizations - NP-47797 Create new trancated field(s) for contributors

Add new field, contributorsPreview, in index document. This will be used in the search response, instead of the contributors list (Ref 413-responses due to too many contributors)

- Extract `contributorsPreview` from expanded resource
- Added fallback logic in cases where expanded resource might not contain `contributorsPreview`

### DIFF
--- a/index-handlers/src/main/java/no/sikt/nva/nvi/index/utils/NviCandidateIndexDocumentGenerator.java
+++ b/index-handlers/src/main/java/no/sikt/nva/nvi/index/utils/NviCandidateIndexDocumentGenerator.java
@@ -82,7 +82,7 @@ public final class NviCandidateIndexDocumentGenerator {
     private static final TypeReference<Map<String, String>> TYPE_REF =
         new TypeReference<>() {
         };
-    private static final int DEFAULT_CONTRIBUTORS_PREVIEW_LIMIT = 5;
+    private static final int FALLBACK_CONTRIBUTORS_PREVIEW_LIMIT = 5;
     private final OrganizationRetriever organizationRetriever;
     private final JsonNode expandedResource;
     private final Candidate candidate;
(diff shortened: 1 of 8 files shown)


index-handlers/src/test/java/no/sikt/nva/nvi/index/IndexDocumentHandlerTest.java
```diff
@@ -328,6 +327,29 @@ void shouldExtractOptionalPrintIssnFromExpandedResource(ChannelType channelType,
...
+        var expectedIndexDocument = setUpExistingResourceInS3AndGenerateExpectedDocument(candidate).indexDocument();
+        var event = createEvent(candidate.getIdentifier());
+        mockUriRetrieverOrgResponse(candidate);
+        handler.handleRequest(event, CONTEXT);
+        var actualIndexDocument = parseJson(s3Writer.getFile(createPath(candidate))).indexDocument();
+        assertEquals(expectedIndexDocument, actualIndexDocument);
+    }
+
+    @Test
+    void shouldSetContributorsPreviewToFiveFirstContributorsIfExpandedResourceIsMissingContributorsPreview() {
```
brinxmat, line 342: Might incidentally want to test what happens when the magic node is an empty array, null, etc.

index-handlers/src/main/java/no/sikt/nva/nvi/index/utils/NviCandidateIndexDocumentGenerator.java
```diff
@@ -80,6 +82,7 @@ public final class NviCandidateIndexDocumentGenerator {
     private static final TypeReference<Map<String, String>> TYPE_REF =
         new TypeReference<>() {
         };
+    private static final int FALLBACK_CONTRIBUTORS_PREVIEW_LIMIT = 5;
```
sondrev, line 85: why not 10, as publication-api uses?

nvi-commons/src/main/java/no/sikt/nva/nvi/common/utils/JsonUtils.java
```diff
@@ -34,7 +34,7 @@ public static URI extractId(JsonNode jsonNode) {
     }
 
     public static boolean isNodePresent(JsonNode node, String jsonPointer) {
-        return !node.at(jsonPointer).isMissingNode();
+        return isNotMissingNode(node.at(jsonPointer));
```
brinxmat, line 37: Maybe use this method directly?

brinxmat (approved):

PR #391, opened 2024-10-18
Np 47912 Index document, publication abstract

monaullah:
Jira task: NP-47912 Search publication abstract

- Add `publicationDetails.abstract` to nvi index document. Extract optional publication abstract from expanded resource
- Add `abstract` to `searchTerm` query

### DIFF
--- a/index-handlers/src/main/java/no/sikt/nva/nvi/index/model/document/PublicationDetails.java
+++ b/index-handlers/src/main/java/no/sikt/nva/nvi/index/model/document/PublicationDetails.java
@@ -1,6 +1,7 @@
 package no.sikt.nva.nvi.index.model.document;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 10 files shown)


index-handlers/src/main/java/no/sikt/nva/nvi/index/model/document/PublicationDetails.java
```diff
@@ -29,6 +31,7 @@ public static final class Builder {
         private String identifier;
         private String type;
         private String title;
+        private String abstractText;
```
truhacevkir, line 34: Maybe simply abstract?

monaullah: `abstract` is a reserved keyword in Java, and can't be used as variable name

monaullah: That's why i set the JsonProperty as "abstract": `@JsonProperty("abstract")`

LarsV123 (approved):

PR #392, opened 2024-10-18
Np 47927 One search hit for unique identifiers

monaullah:
Jira task: NP-47927 Search for unique identifiers should only give one hit

Thanks for the help @StigNorland 😄 

- Modify mulitimatch query, add operator `And` and type `CrossFields`


StigNorland (approved):
👍

PR #399, opened 2024-11-07
Np 48035 Candidate.upsert() void

monaullah:
Jira task: NP-48035 - CloudWath Alarm: NVI Persisting candidate failed

Due to:
```
java.lang.NullPointerException: Cannot invoke \"no.sikt.nva.nvi.common.db.CandidateDao.identifier()\" because \"candidateDao\" is null\n\t
at no.sikt.nva.nvi.common.service.model.Candidate.createCandidate(Candidate.java:441)\n\tat 
no.sikt.nva.nvi.common.service.model.Candidate.upsert(Candidate.java:141)\n\tat 
...
```

Anlalysis: Persisting the candidate has not _actually_ failed, but _fetching_ the candidate immediately after creating it in some very few cases returns `null`. I think this is due to the eventual consistency model of DynamoDB.

Discussed following alternatives with @LarsV123:
1. Use `GetItemEnhancedRequest` with `consistentRead` set to `true` when fetching the candidate immediately after creating it. This ensures that a read immediately following a write will always return the latest data, but is slower and more expensive.
2. If fetching the candidate immediately after creating it returns `null`, retry?
3. The return value of method `Candidate.upsert()` is actually never used (in production code) - therefore: Change return type to void?

Option 3. is implemented in this PR, but feel free to disagree or suggest other alternatives.

### DIFF
--- a/event-handlers/src/main/java/no/sikt/nva/nvi/events/persist/UpsertNviCandidateHandler.java
+++ b/event-handlers/src/main/java/no/sikt/nva/nvi/events/persist/UpsertNviCandidateHandler.java
@@ -14,7 +14,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import no.sikt.nva.nvi.common.db.CandidateRepository;
-import no.sikt.nva.nvi.common.db.PeriodRepository;
 import no.sikt.nva.nvi.common.queue.NviQueueClient;
 import no.sikt.nva.nvi.common.queue.QueueClient;
 import no.sikt.nva.nvi.common.service.model.Candidate;
@@ -36,22 +35,18 @@ public class UpsertNviCandidateHandler implements RequestHandler<SQSEvent, Void>
     private static final String UPSERT_CANDIDATE_DLQ_QUEUE_URL = "UPSERT_CANDIDATE_DLQ_QUEUE_URL";
     private static final String UPSERT_CANDIDATE_FAILED_MESSAGE = "Failed to upsert candidate for publication: {}";
     private final CandidateRepository repository;
-    private final PeriodRepository periodRepository;
-
     private final QueueClient queueClient;
     private final String dlqUrl;
 
     @JacocoGenerated
     public UpsertNviCandidateHandler() {
-        this(new CandidateRepository(defaultDynamoClient()), new PeriodRepository(defaultDynamoClient()),
-             new NviQueueClient(), new Environment());
+        this(new CandidateRepository(defaultDynamoClient()), new NviQueueClient(), new Environment());
     }
 
-    public UpsertNviCandidateHandler(CandidateRepository repository, PeriodRepository periodRepository,
+    public UpsertNviCandidateHandler(CandidateRepository repository,
                                      QueueClient queueClient,
                                      Environment environment) {
         this.repository = repository;
-        this.periodRepository = periodRepository;
         this.queueClient = queueClient;
         this.dlqUrl = environment.readEnv(UPSERT_CANDIDATE_DLQ_QUEUE_URL);
     }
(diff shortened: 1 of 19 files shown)


nvi-commons/src/test/java/no/sikt/nva/nvi/common/service/CandidateTest.java
```diff
@@ -97,23 +98,35 @@ class CandidateTest extends LocalDynamoTest {
...
     private static final String API_DOMAIN = ENVIRONMENT.readEnv("API_HOST");
     public static final URI CONTEXT_URI = UriWrapper.fromHost(API_DOMAIN).addChild(BASE_PATH, "context").getUri();
+    private static final URI HARDCODED_INSTITUTION_ID = URI.create(
+        "https://example"
+        + ".org/hardCodedInstitutionId");
     private CandidateRepository candidateRepository;
     private PeriodRepository periodRepository;
 
     public static Stream<Arguments> candidateResetCauseProvider() {
+        //TODO: Test institution changed
```
monaullah, line 108: Added a TODO there when i noticed bug, created jira task: [Remove approval for institution A when creator affiliation changes to institution B](https://sikt.atlassian.net/browse/NP-48041)

brinxmat (approved):
Does this change imply changes in OpenApi spec, or more generally status codes?

monaullah:
> Does this change imply changes in OpenApi spec, or more generally status codes?

No, `UpsertNviCandidateHandler` is not a `ApiGatewayHandler`

truhacevkir (approved):

event-handlers/src/main/java/no/sikt/nva/nvi/events/persist/UpsertNviCandidateHandler.java
```diff
@@ -36,22 +35,18 @@ public class UpsertNviCandidateHandler implements RequestHandler<SQSEvent, Void>
     private static final String UPSERT_CANDIDATE_DLQ_QUEUE_URL = "UPSERT_CANDIDATE_DLQ_QUEUE_URL";
     private static final String UPSERT_CANDIDATE_FAILED_MESSAGE = "Failed to upsert candidate for publication: {}";
     private final CandidateRepository repository;
-    private final PeriodRepository periodRepository;
-
     private final QueueClient queueClient;
     private final String dlqUrl;
 
     @JacocoGenerated
     public UpsertNviCandidateHandler() {
```
LarsV123, line 42: Not related to this PR, but:
Do we actually need this constructor? It isn't used and the tests still pass if I remove it.

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/Candidate.java
```diff
@@ -135,12 +135,10 @@ public static Candidate fetch(FetchCandidateRequest request, CandidateRepository
...
+    public static void upsert(UpsertCandidateRequest request, CandidateRepository repository) {
         var optionalCandidate = repository.findByPublicationId(request.publicationId());
-        return optionalCandidate.map(
-                candidateDao -> handleExistingCandidate(request, repository, periodRepository, candidateDao))
-                   .orElseGet(() -> Optional.of(createCandidate(request, repository, periodRepository)));
+        optionalCandidate.ifPresentOrElse(candidateDao -> updateExistingCandidate(request, repository, candidateDao),
+                                          () -> createCandidate(request, repository));
     }
 
     public static Optional<Candidate> updateNonCandidate(UpdateNonCandidateRequest request,
```
LarsV123, line 144: Should this return `void` too? It doesn't seem like the return value is used in production code here either.

LarsV123 (approved):

event-handlers/src/main/java/no/sikt/nva/nvi/events/persist/UpsertNviCandidateHandler.java
monaullah: Yes, it's the default constructor used to initialize the Lambda in AWS

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/Candidate.java
monaullah: Probably, future PR?

PR #400, opened 2024-11-08
Rename and tidy in queues and dlqs

monaullah:
Jira task: NP-48059 CloudWath Alarm: NVI Persisting candidate failed (TransactionConflict)

Noticed that dlq redrive was not configured correctly for `UpsertCandidateDLQ`, so I took a "tidy" round for queue names (making them more intuitive in my opinion) and configured `UpsertCandidateDLQ` as target in redrive policy for `EvaluatedCandidateQueue`

Changes:
- Change queue name `PersistedResourceQueue` to `ResourceEvaluationQueue`
- Change queue name `NewCandidateQueue` to `EvaluatedCandidateQueue` (since this queue contains both `NviCandidate`s and `NonNviCandidate`s)
- Remove `NewCandidateDLQ` and set `EvaluateNviCandidateHandler` DLQ to `EvaluationDLQ`
- Set `UpsertCandidateDLQ` as redrive target for `EvaluatedCandidateQueue`
- Added `QueuePersistedResourceDLQ` for `QueuePersistedResourceHandler`

This change does not fix the jira bug directly, but allows automatic retries immediately after in cases like TransactionConflict


LarsV123 (approved):
Will this cause the queues to be deleted and recreated? If so, is that a problem?

monaullah:
> Will this cause the queues to be deleted and recreated? If so, is that a problem?

I tested this in sandbox. Yes, this causes the queues to be deleted and recreated but this should not be an issue

PR #404, opened 2024-11-15
NP-48059 - Update candidate and remove approvals in one transaction

monaullah:
Jira task: NP-48059 CloudWath Alarm: NVI Persisting candidate failed (TransactionConflict)

Not sure if this is related to the bug mentioned above, but it is better to do this update in one transaction instead of two in method `updateCandidateAndRemovingApprovals` in `CandidateRepository`.

### DIFF
--- a/nvi-commons/src/main/java/no/sikt/nva/nvi/common/db/CandidateRepository.java
+++ b/nvi-commons/src/main/java/no/sikt/nva/nvi/common/db/CandidateRepository.java
@@ -121,6 +121,12 @@ public class CandidateRepository extends DynamoRepository {
         return candidateTable.getItem(candidate);
     }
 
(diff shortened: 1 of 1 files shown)


nvi-commons/src/main/java/no/sikt/nva/nvi/common/db/CandidateRepository.java
```diff
@@ -121,6 +121,12 @@ public CandidateDao create(DbCandidate dbCandidate, List<DbApprovalStatus> appro
         return candidateTable.getItem(candidate);
     }
 
+    public void updateCandidate(CandidateDao candidate) {
```
monaullah, line 124: No change here, just moved this up to keep the update-methods together

LarsV123 (approved):

PR #418, opened 2025-01-14
Np 48266 adding authorization to fetch candidate by pub id endpoint

truhacevkir:
Adding back authentication to `FetchNviCandidateByPublicationIdHandler`.
Using same validation rule as in `FetchNviCandidateHandler`.


nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/dto/CandidateDto.java
```diff
@@ -32,13 +31,6 @@ public static Builder builder() {
         return new Builder();
     }
 
-    //TODO: To remove when frontend has been updated to use "period" property
-    @Deprecated
-    @JsonGetter("periodStatus")
-    public PeriodStatusDto getPeriodStatus() {
-        return this.period;
-    }
-
```
truhacevkir, line 41: Frontend is migrated, removing todo and not used field.

torbjokv (approved):

nvi-rest/src/main/java/no/sikt/nva/nvi/rest/fetch/FetchNviCandidateByPublicationIdHandler.java
```diff
@@ -48,6 +55,8 @@ protected CandidateDto processInput(Void input, RequestInfo requestInfo, Context
                    .map(identifier -> Candidate.fetchByPublicationId(() -> identifier, candidateRepository,
                                                                      periodRepository))
                    .map(this::checkIfApplicable)
+                   .map(candidate -> validateViewingScope(viewingScopeValidator, RequestUtil.getUsername(requestInfo),
```
brinxmat, line 58: Is there a reason to not do this _much_ earlier?

truhacevkir: It would be better to verify before fetching candidate, but we need organizations from NVI candidate to check if user belongs to one of organizations, so it should be done after we have fetched candidate from database.

LarsV123 (approved):

PR #421, opened 2025-01-17
[np-48171] Add validation check to approvals for unverified creators

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-48171

This adds basic validation to ensure that institutions cannot approve/reject NVI candidates if they have affiliated unverified creators. The placement of the logic is sub-optimal because the necessary metadata isn't persisted (yet), and doing it properly would likely require significant refactoring.


nvi-rest/src/main/java/no/sikt/nva/nvi/rest/upsert/NviStatusRequest.java
```diff
@@ -11,6 +12,12 @@ public record NviStatusRequest(UUID candidateId,
                                ApprovalStatus status,
                                String reason) {
 
+    public void validate() {
+        if (isNull(candidateId) || isNull(institutionId) || isNull(status)) {
+            throw new IllegalArgumentException("Missing required fields");
```
LarsV123, line 17: Adding this as a sanity check, because I assume it should not be possible to update the status of an NVI candidate without specifying these fields.

nvi-rest/src/test/java/no/sikt/nva/nvi/rest/upsert/UpdateNviCandidateStatusHandlerTest.java
```diff
@@ -167,31 +225,33 @@ void shouldUpdateApprovalStatus(ApprovalStatus oldStatus, ApprovalStatus newStat
...
-        var request = createRequest(candidate.getIdentifier(), institutionId, newStatus);
+        var request = createRequest(candidate.getIdentifier(), DEFAULT_TOP_LEVEL_INSTITUTION_ID, newStatus);
         handler.handleRequest(request, output, context);
         var response = GatewayResponse.fromOutputStream(output, CandidateDto.class);
         var candidateResponse = response.getBodyObject(CandidateDto.class);
         var actualApproval = candidateResponse.approvals().getFirst();
         assertThat(actualApproval.finalizedBy(), is(nullValue()));
         assertThat(actualApproval.finalizedDate(), is(nullValue()));
-        assertThat(actualApproval.status(), is(equalTo(ApprovalStatusDto.NEW)));
+        assertThat(actualApproval.status(), is(equalTo(ApprovalStatusDto.PENDING)));
```
LarsV123, line 240: Note that this changes the assert from `status=NEW` to `status=PENDING`. I believe it was an error in the test setup that made it possible to get `status=NEW`, which is now fixed (by always including the institution ID in the update request).

brinxmat (approved):

PR #427, opened 2025-01-22
fix: Replace objectMapper

LarsV123:
This is intended to fix an issue with parsing `CandidateDao` after changes to the data model, by replacing the object mapper used throughout the project. The `dynamoObjectMapper` used previously ignored empty fields (i.e. null values), which I believe caused issues with parsing certain values.

### DIFF
--- a/event-handlers/src/test/java/no/sikt/nva/nvi/events/evaluator/dto/ContributorDtoTest.java
+++ b/event-handlers/src/test/java/no/sikt/nva/nvi/events/evaluator/dto/ContributorDtoTest.java
@@ -20,7 +20,6 @@ class ContributorDtoTest {
               "verificationStatus": "NotVerified",
               "additionalIdentifiers": []
           },
-          "affiliations": [],
           "role": {
               "type": "Creator"
           },
--- a/event-handlers/src/test/java/no/sikt/nva/nvi/events/evaluator/dto/ContributorDtoTest.java
+++ b/event-handlers/src/test/java/no/sikt/nva/nvi/events/evaluator/dto/ContributorDtoTest.java
@@ -6,11 +6,34 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
-import org.junit.jupiter.api.Test;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Named;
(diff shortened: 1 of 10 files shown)


LarsV123:
Tested this in `dev` by running `BatchReEvaluateNviCandidates`, seeing no parsing errors. Appears to work and fix the parsing issue.

ketilaa (approved):

event-handlers/src/test/java/no/sikt/nva/nvi/events/evaluator/dto/ContributorDtoTest.java
```diff
@@ -20,7 +20,6 @@ class ContributorDtoTest {
               "verificationStatus": "NotVerified",
               "additionalIdentifiers": []
           },
-          "affiliations": [],
```
brinxmat, line 23: This is to provoke the error? We could do destruction testing on stupid cases directly with a null and empty source in a parameterized test

brinxmat (approved):

event-handlers/src/test/java/no/sikt/nva/nvi/events/evaluator/dto/ContributorDtoTest.java
LarsV123: Indeed, just a lazy way to test the actual scenario I had. Should parameterize it to test both cases.

nvi-commons/src/test/java/no/sikt/nva/nvi/common/utils/DynamoDbUtilsTest.java
```diff
@@ -106,4 +61,49 @@ private static DynamodbStreamRecord setupStreamRecordWithFieldWithNullValue() {
...
+  }
+
+  @Test
+  void shouldSerializeWithoutLossOfData() throws Exception {
+    var originalStreamRecord =
+        dynamodbRecordWithStreamRecord(streamRecordWithOldAndNewImage(UUID.randomUUID()));
+    var streamRecordAsString = dtoObjectMapper.writeValueAsString(originalStreamRecord);
+    var regeneratedStreamRecord = DynamoDbUtils.toDynamodbStreamRecord(streamRecordAsString);
+    assertEquals(originalStreamRecord, regeneratedStreamRecord);
+  }
```
LarsV123, line 108: These methods were re-arranged automatically, I don't know why.

brinxmat (approved):

PR #431, opened 2025-01-29
[np-48420] refactor: Use base class for test setup

LarsV123:
Adds an abstract base class for duplicated test setup. This is necessary to support new tests for this ticket.

### DIFF
--- a/nvi-rest/src/test/java/no/sikt/nva/nvi/rest/BaseCandidateRestHandlerTest.java
+++ b/nvi-rest/src/test/java/no/sikt/nva/nvi/rest/BaseCandidateRestHandlerTest.java
@@ -49,46 +49,58 @@ import no.unit.nva.testutils.HandlerRequestBuilder;
 import nva.commons.apigateway.AccessRight;
 import nva.commons.apigateway.ApiGatewayHandler;
 import nva.commons.apigateway.GatewayResponse;
+import nva.commons.core.JacocoGenerated;
 import org.apache.hc.core5.http.ContentType;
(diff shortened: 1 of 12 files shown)


nvi-rest/src/test/java/no/sikt/nva/nvi/rest/BaseCandidateRestHandlerTest.java
```diff
@@ -0,0 +1,298 @@
...
+  protected static UnverifiedNviCreatorDto setupUnverifiedCreator(
+      String name, Collection<URI> affiliations, URI topLevelInstitutionId) {
+    affiliations.forEach(
+        affiliation ->
+            mockOrganizationResponseForAffiliation(
+                topLevelInstitutionId, affiliation, mockUriRetriever));
+    return new UnverifiedNviCreatorDto(name, List.copyOf(affiliations));
+  }
+
+  // Not used (yet)
```
brinxmat, line 247: In general, mark these with `// TODO: ` as these are picked up as warnings in the IDE, which is a bit more of a reminder to do what we intend, even if we come back to the task after a month of fever dreams

brinxmat (approved):

nvi-rest/src/test/java/no/sikt/nva/nvi/rest/BaseCandidateRestHandlerTest.java
LarsV123: Good point! These are being removed in another PR today, but I'll keep it in mind.

PR #439, opened 2025-02-01
build: Upgrade to DynamoDBLocal 2.5.4

LarsV123:


buildSrc/src/main/groovy/nva.nvi.rootplugin.gradle
```diff
@@ -7,7 +7,6 @@ plugins {
 repositories {
     mavenCentral()
     maven { url 'https://jitpack.io' }
```
brinxmat, line 9: Can you try building without jitpack…I don't think it is needed

brinxmat (approved):
Probably a happier reviewer if you merge commits into the other branch, rather than main so I only see the current changes.

PR #453, opened 2025-02-27
[np-48641] fix: Handle candidates without verified creators

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-48641

Fixes issues with generating Excel reports from index documents with unverified creators by mapping missing values to zero.
This scenario should possibly be prevented before the documents are created, but that would require a full re-indexing and more work.

### DIFF
--- a/index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/FetchInstitutionReportHandlerTest.java
+++ b/index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/FetchInstitutionReportHandlerTest.java
@@ -5,9 +5,10 @@ import static com.google.common.net.HttpHeaders.CONTENT_TYPE;
 import static com.google.common.net.MediaType.ANY_APPLICATION_TYPE;
 import static com.google.common.net.MediaType.MICROSOFT_EXCEL;
 import static com.google.common.net.MediaType.OOXML_SHEET;
+import static java.util.Collections.emptyList;
 import static java.util.Objects.nonNull;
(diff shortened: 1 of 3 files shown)


index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/FetchInstitutionReportHandlerTest.java
```diff
@@ -314,15 +335,14 @@ void shouldNotContainDataForOtherInstitutions() throws IOException {
...
+    var indexDocument = indexDocumentMissingApprovals(CURRENT_YEAR, topLevelCristinOrg);
     when(openSearchClient.search(any()))
         .thenReturn(aggregationResponse(1))
         .thenReturn(createSearchResponse(indexDocument));
     var appender = LogUtils.getTestingAppender(NviCandidateIndexDocument.class);
     handler.handleRequest(
         requestWithMediaType(MICROSOFT_EXCEL.toString(), topLevelCristinOrg), output, CONTEXT);
-    assertTrue(appender.getMessages().contains(indexDocument.id().toString()));
+    assertTrue(appender.getMessages().contains(indexDocument.identifier().toString()));
   }
```
LarsV123, line 346: Note that this test used to trigger a different error that did not match the test name and can no longer be easily triggered. This fixes this test case to match the name, but the original exception in `no.sikt.nva.nvi.index.model.document.NviCandidateIndexDocument#generateRow` is now untested.

brinxmat (approved):

PR #454, opened 2025-02-27
refactor: Improve use of test fixtures

LarsV123:
Expands and improves the use of test fixtures in various test classes.

Main changes:
- Make `LocalDynamoTestSetup` purely static helper methods instead of a base class that tests extend.
- Use `TestScenario` in more test classes to simplify setup of common fake/mocked services.


event-handlers/src/test/java/no/sikt/nva/nvi/events/batch/EventBasedBatchScanHandlerTest.java
```diff
@@ -80,16 +80,15 @@ class EventBasedBatchScanHandlerTest extends LocalDynamoTestSetup {
   private static final int PAGE_SIZE = 4;
   private EventBasedBatchScanHandler handler;
   private ByteArrayOutputStream output;
-  private Context context;
+  private static final Context CONTEXT = mock(Context.class);
```
brinxmat, line 83: We should probably have done the job needed to make FakeContext work, but fair here.

brinxmat (approved):

PR #455, opened 2025-03-07
[np-48616] feat:  Add Cucumber testing framework

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-48616

Changes:
- feat: Add skeleton code and minimal example of Cucumber testing, so that future work can include tests in this format.
- fix: Disable telemetry for DynamoDBLocal in tests, reducing test execution time significantly.
- fix: Correct typo in error message (`UnverifiedCreatorFromOrganizationProblem.java`)


.run/Template Cucumber Java.run.xml
LarsV123, line 1: This template is used by IntelliJ when running Cucumber tests from the GUI, which bypasses the usual Gradle flow that sets up environments variables for other tests.

ketilaa (approved):

torbjokv (approved):

PR #456, opened 2025-03-11
[np-48710] fix: Use year param in search

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-48710

This changes candidate search to filter by reporting period year instead of publication year.
These should be identical, but some existing data has a mismatch between these fields which caused the candidate to show up for the wrong year in the curator workflow.


index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/SearchNviCandidatesHandlerTest.java
```diff
@@ -460,6 +462,23 @@ void shouldExcludeContributorsInSearchResponse() throws IOException {
         .search(argThat(argument -> argument.excludeFields().equals(expectedExcludeFields)));
   }
 
+  @Test
+  void shouldSearchByReportingPeriodWhenYearParamIsSet() throws IOException {
```
ketilaa, line 466: Seems to me like this test the `id` of the search result, not that we are actually searching for reporting year, or?

ketilaa (approved):

index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/SearchNviCandidatesHandlerTest.java
LarsV123: The `id` field in the response is the query string that produces this response (which should be what we actually searched for). That's all we can really test in this class, because the actual OpenSearch client is mocked and integration tests done separately.

PR #472, opened 2025-04-14
[np-48689] perf: Simplify PointService

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-48689

This refactors `PointService` to use `PublicationDto` instead of re-creating the necessary data using `JsonNode` and extra networks calls. This in effect eliminates network calls checking for top-level organizations of Contributors during the evaluation stage. Note that we still check the NVI status of top-level organizations here and that the old logic is duplicated in the indexing stage, which must be addressed separately.

It also fixes a bug in the "is this an international collaboration" part of the SPARQL query, as this did not previously check that the affiliation was for a `Creator`.

This PR shrinks `PointService` to the point that it perhaps does not make sense to keep it as a separate "service class", but it is left for now to keep the PR short and readable.


nvi-commons/src/main/resources/publication_query.sparql
```diff
@@ -178,4 +169,15 @@ WHERE {
       }
     }
   }
+
```
brinxmat, line 172: Denne ser annerledes ut da den andre omhandlet affiliations

brinxmat (approved):

PR #478, opened 2025-04-16
build: Remove CheckStyle

LarsV123:
Removing CheckStyle from the project because almost all the rules overlap with other, better tools. Spotless handles formatting (which is most of CheckStyle) while ErrorProne and PMD provide static analysis.


brinxmat (approved):
Can you add an explanation why you're removing it — as an aid to later reviewers and commit trawling

PR #479, opened 2025-04-23
fix: Handle parsing of large documents

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49010

This replaces some `OPTIONAL` blocks with `UNION`, which significantly increases performance. The synthetic test case has similar performance (roughly 1 second per 1000 contributors) as the "production" documents that were causing time-outs.


event-handlers/src/test/java/no/sikt/nva/nvi/events/evaluator/EvaluateNviCandidateWithSyntheticDataTest.java
LarsV123, line 1: This test suite just has one test now and seems pointless, but the intention is to split up the very messy `EvaluateNviCandidateHandlerTest` file and re-do some of those tests in this format.

brinxmat (approved):

PR #485, opened 2025-05-14
feat: Add integration testing

LarsV123:
Changes:
- Adds a new Gradle subproject for integration tests with a minimal example of Cucumber tests covering multiple handlers. This is necessary setup for future changes and will allow us to merge many existing tests.
- Adds a new `FakeEnvironment` utility to set up fake environment variables for tests. This is a "proof of concept" and not fully used yet, but the goal is to remove all environment variables defined in top-level `build.gradle` files etc.
- Includes some refactoring/clean-up of test util methods.


event-handlers/src/test/java/no/sikt/nva/nvi/events/evaluator/EvaluateNviCandidateWithSyntheticDataTest.java
```diff
@@ -45,13 +43,12 @@ void setup() {
   // The parser should be able to handle documents with 10 000 contributors in 30 seconds.
   // This test case is a bit more generous because the GitHub Actions test runner is underpowered.
   @ParameterizedTest
-  @Timeout(value = 30, unit = TimeUnit.SECONDS)
+  @Timeout(value = 45, unit = TimeUnit.SECONDS)
```
LarsV123, line 46: This is an unrelated change, but the GitHub actions runner is so slow that this test was already close to the limit and has now started breaking it. We should revisit this at some point...

brinxmat (approved):

PR #488, opened 2025-05-20
[np-49176] fix: Simplify data in update events

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49176

This simplifies the data being sent in the indexing event flow in order to avoid issues with large documents on SQS queues. The actual functionality should be unchanged, but the PR touches on many different parts of the system and test setup.

Main points:

- A new `DynamoDbChangeMessage` record is added that only contains the information relevant for the indexing process. The stream events from DynamoDB are converted to this immediately in `DynamoDbEventToQueueHandler` and the new model passed through the rest of the event flow.
- Some existing test suites had to be rewritten entirely, and I took the opportunity to do this as integration tests (testing both handlers in the same suite). This will make merging the handlers easier later on.
- Some existing test cases were deleted, as the scenarios tested were no longer possible.


integration-tests/src/test/java/no/sikt/nva/nvi/events/DynamoDbUpdateEventTest.java
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  @Test
+  void shouldSendMessageToDlqIfSendingBatchFails() {
+    var candidateIdentifier = randomUUID();
+    var dynamoDbEvent = createCandidateEvent(candidateIdentifier, OperationType.MODIFY, true);
+
+    // Rig the sharedQueueClient to throw an exception when sending a message to the queue
+    sharedQueueClient = spy(FakeSqsClient.class);
+    doThrow(SqsException.class).when(sharedQueueClient).sendMessageBatch(any(), any());
```
brinxmat, line 142: Since FakeSqsClient is "our" code, should we not be able to throw when dispatching without this mockery?

integration-tests/src/test/java/no/sikt/nva/nvi/events/DynamoDbUpdateEventTest.java
```diff
@@ -0,0 +1,367 @@
...
+  private List<String> extractBatchEntryMessageBodiesAtIndex(int index) {
+    return sharedQueueClient.getSentBatches().get(index).entries().stream()
+        .map(SendMessageBatchRequestEntry::messageBody)
+        .toList();
+  }
+
+  private static Stream<Arguments> candidateEventProvider() {
+    var identifier = randomUUID();
+    return Stream.of(
+        argumentSet(
```
brinxmat, line 285: Haven't run the tests, so I don't know — how is this argument-set-name displayed in the IDE/test results?

nvi-commons/src/main/java/no/sikt/nva/nvi/common/queue/DynamoDbChangeMessage.java
```diff
@@ -0,0 +1,39 @@
...
+ * database. This is used for records that either are or are directly related to a Candidate. This
+ * is used to avoid sending the entire original DynamoDB event message, which can be too big to fit
+ * on a queue.
+ *
+ * @param candidateIdentifier the identifier of the related candidate
+ * @param entryType the type of entry being changed, which is not necessarily the same as the type
+ *     of the DAO
+ * @param operationType the type of DynamoDB operation that triggered the event
+ */
+public record DynamoDbChangeMessage(
```
brinxmat, line 21: Amazing that creating a structure that actually represents the thing you're trying to manipulate with necessary accessors for the processing you intend to do makes things easier. Astounded. Have fallen off chair. I am sure you are doing something illegal, which is why I have not seen this before.

nvi-commons/src/testFixtures/java/no/sikt/nva/nvi/common/QueueServiceTestUtils.java
```diff
@@ -1,108 +1,98 @@
...
+    var dbMessage = createDbChangeMessage(candidateIdentifier);
+    return createEvent(dbMessage);
   }
 
-  public static SQSEvent createEventWithDynamodbRecords(List<DynamodbStreamRecord> streamRecords) {
-    var sqsEvent = new SQSEvent();
-    var messages = streamRecords.stream().map(QueueServiceTestUtils::createMessage).toList();
-    sqsEvent.setRecords(messages);
-    return sqsEvent;
+  public static SQSEvent createEvent(List<UUID> candidateIdentifiers) {
```
brinxmat, line 47: Collection<UUID>

event-handlers/src/main/java/no/sikt/nva/nvi/events/db/DynamoDbEventToQueueHandler.java
```diff
@@ -56,17 +71,51 @@ public Void handleRequest(DynamodbEvent input, Context context) {
...
+        .toList();
   }
 
-  private static String writeAsJsonString(DynamodbStreamRecord record) {
-    return attempt(() -> dtoObjectMapper.writeValueAsString(record)).orElseThrow();
+  private static DynamoDbChangeMessage mapToUpdateMessage(DynamodbStreamRecord streamRecord) {
+    var operationType = OperationType.fromValue(streamRecord.getEventName());
+    var entryType = getEntryType(streamRecord);
+    if (entryType.shouldBeProcessedForIndexing()) {
+      var recordIdentifier = UUID.fromString(extractField(streamRecord, IDENTIFIER_FIELD));
```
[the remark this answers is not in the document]
LarsV123: Added a validation step to check for this, so we can handle the exception here.

PR #489, opened 2025-05-22
[np-49204] fix: Avoid sending empty message batches

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49204

### DIFF
--- a/event-handlers/src/main/java/no/sikt/nva/nvi/events/db/DynamoDbEventToQueueHandler.java
+++ b/event-handlers/src/main/java/no/sikt/nva/nvi/events/db/DynamoDbEventToQueueHandler.java
@@ -71,14 +71,6 @@ public class DynamoDbEventToQueueHandler implements RequestHandler<DynamodbEvent
     return null;
   }
 
-  private static List<String> mapToUpdateMessages(List<DynamodbStreamRecord> records) {
-    return records.stream()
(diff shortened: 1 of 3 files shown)


event-handlers/src/main/java/no/sikt/nva/nvi/events/db/DynamoDbEventToQueueHandler.java
```diff
@@ -116,9 +108,13 @@ private static String writeAsJsonString(DynamoDbChangeMessage updateMessage) {
   }
 
   private void splitIntoBatchesAndSend(DynamodbEvent input) {
-    splitIntoBatches(input.getRecords())
-        .map(DynamoDbEventToQueueHandler::mapToUpdateMessages)
-        .forEach(this::sendBatch);
+    var messages =
+        input.getRecords().stream()
+            .map(DynamoDbEventToQueueHandler::mapToUpdateMessage)
```
brinxmat, line 113: Just a thought, we expect distinct, right?

LarsV123: Yes, but I don't think it matters much (a duplicate message should just be extra load). I can add `distinct` here just in case though.

brinxmat (approved):

PR #491, opened 2025-05-22
[np-49197] fix: Handle contributors with multiple roles

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49197

Adds handling of documents where the same contributors is listed multiple times with different roles.
For simplicity, this is done by discarding all contributors that are not creators, as they are irrelevant here.
If non-creator contributors become relevant, we will have to handle this differently.


nvi-commons/src/test/java/no/sikt/nva/nvi/common/dto/ContributorDtoTest.java
```diff
@@ -4,7 +4,6 @@
 import static no.sikt.nva.nvi.common.examples.ExampleContributors.EXAMPLE_2_CONTRIBUTOR_1;
 import static no.sikt.nva.nvi.common.examples.ExampleContributors.EXAMPLE_2_CONTRIBUTOR_2;
 import static no.sikt.nva.nvi.common.examples.ExampleContributors.EXAMPLE_2_CONTRIBUTOR_3;
```
brinxmat, line 6: Is there a reason for these changes?

brinxmat (approved):

PR #493, opened 2025-05-26
fix: Use null-check for migrated fields

LarsV123:
Fixes null-pointer exceptions on records that are not migrated yet.

### DIFF
--- a/nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/PublicationDetails.java
+++ b/nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/PublicationDetails.java
@@ -85,18 +85,11 @@ public record PublicationDetails(
         builder()
             .withId(dbCandidate.publicationId())
             .withPublicationBucketUri(dbCandidate.publicationBucketUri())
-            .withTitle(dbDetails.title())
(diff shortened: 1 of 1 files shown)


nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/PublicationDetails.java
```diff
@@ -85,18 +85,11 @@ public static PublicationDetails from(CandidateDao candidateDao) {
...
             .withIsApplicable(dbCandidate.applicable())
             .withPublicationChannel(PublicationChannel.from(candidateDao))
             .withNviCreators(nviCreators)
-            .withCreatorCount(dbDetails.contributorCount())
-            .withTopLevelOrganizations(topLevelOrganizations)
-            .withModifiedDate(dbDetails.modifiedDate());
+            .withTopLevelOrganizations(topLevelOrganizations);
 
     if (nonNull(dbDetails)) {
       return getPublicationDetailsWithMigratedFields(dbDetails, builder);
```
LarsV123, line 95: The "removed" fields are set here if the record is migrated, otherwise we do not have them anyway.

brinxmat (approved):

PR #495, opened 2025-06-02
removing approvals for the same institution when migrating nvi candidates from Cristin

truhacevkir:
When migrating nvi candidates from Cristin, we use CristinLocales to construct approvals. It shows up that nvi candidate in Cristin can contain multiple "approvals" pointing to the same institution. Removing those duplicates when importing them, as today they are failing on condition when persisting to database, cause institutionId for approval is primary key in DynamoDB. 

`Transaction request cannot include multiple operations on one item (Service: DynamoDb, Status Code: 400, Request ID: VOQHVFCPJ25L4D77N1SU5LTF4BVV4KQNSO5AEMVJF66Q9ASUAAJG) (SDK Attempt Count: 1)`


scoreboard-task
LarsV123, line 1: Is this intentional? If so, what is it?

LarsV123 (approved):

brinxmat (approved):

PR #497, opened 2025-06-06
[np-49276] fix: Handle unexpected enum values

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49276

This adds a default value to `InstanceType`, `ChannelType`, and `ScientificValue` enums in order to handle historical records with unexpected values. This means that historical records can be processed and indexed, and the validation of these values now only happens once on the way in (either in the normal evaluation flow or in migration/import flow).


brinxmat (approved):
Tidy

PR #505, opened 2025-06-16
[np-49257] refactor: Address QA feedback

LarsV123:
Quick, separate PR to address suggestions from the previous PR.


nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/Username.java
```diff
@@ -1,6 +1,9 @@
 package no.sikt.nva.nvi.common.service.model;
 
 import static java.util.Objects.isNull;
+import static nva.commons.core.StringUtils.isBlank;
+
+import no.sikt.nva.nvi.common.exceptions.ValidationException;
```
brinxmat, line 6: Does this override `toString`, if it can be overridden, just return the value and make the class final

brinxmat (approved):

PR #506, opened 2025-06-16
adding more transfers to import last bunch of failing candidates

truhacevkir:
Adding more transfers to import last bunch of 11 failing candidates where contributors and approvals does not match.


truhacevkir:
> Out of scope for this PR, but:
> 
> 1. Is the purpose/content of these files documented anywhere?
> 2. Why do we need to duplicate the CSV file in the test resources?

Svar på 1 og 2 er -> latskap :) Men burde absolutt ha skrevet noen linjer om csv'en og blitt kvitt den andre csv test filen :)

PR #520, opened 2025-06-26
[np-49331] docs: Add OpenAPI for planned changes

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49331

Changes:

- Adds query parameters to OpenAPI spec for planned changes:
  - `status`: Filter by one of multiple approval status for current user organization
  - `globalStatus`: Filter by one of multiple global approval statuses
  - `excludeUnassigned`: Exclude candidates with no assigned curator.
  - `filter`: Three new special filters that replace the existing "filters". These cover special requirements that cannot be combined with each other.
- Removes current `filter` options from OpenAPI spec. The endpoint still supports them, but they will be removed when the planned changes are implemented and the frontend switches over.
- Fixes various typos and minor errors in the existing OpenAPI spec
- Improves examples and validation of existing OpenAPI spec

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -337,9 +337,16 @@ paths:
           schema:
             type: string
           description: |
-            Filters by ID of the assigned curator for an organization.
+            Filters by ID of the assigned curator for an organization. By default, candidates with
+            no assigned curator are also included.
             Expected format is `<user_id>@<organization_id>`.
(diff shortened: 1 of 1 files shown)


elkebab (approved):
Dette vil oppfylle alle krav, så vidt jeg kan se 👏 
Starter implementasjon for dette i frontend allerede nå for å raskere kunne verifisere

PR #524, opened 2025-07-01
refactor: Restructure search tests for maintainability

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49374

This is an initial refactoring to make it possible to write tests for the planned changes.

Changes:
- Splits the existing `SearchNviCandidatesHandler` in three:
  - Adds a base class that the other two test suites extend (`SearchNviCandidatesHandlerTestBase`)
  - The existing `SearchNviCandidatesHandlerTest` uses a mocked OpenSearchClient (as before) and some static data, but with test setup and assertions simplified. Hamcrest is removed and replaced with AssertJ.
  - `SearchNviCandidatesHandlerIntegrationTest` uses a TestContainer for OpenSearch, making the tests integration tests. This is set up with nested subclasses for different groups of tests (nearly empty placeholders for now).
- Extracts reusable utility methods from `OpenSearchClientTest` and  `SearchNviCandidatesHandlerTest` to separate classes.

### DIFF
--- a/index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/SearchNviCandidatesHandlerTestBase.java
+++ b/index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/SearchNviCandidatesHandlerTestBase.java
@@ -81,16 +81,6 @@ public class SearchNviCandidatesHandlerTestBase {
     }
   }
 
-  protected Problem handleBadRequest(InputStream request) {
-    try {
-      handler.handleRequest(request, output, CONTEXT);
-      var response = GatewayResponse.fromOutputStream(output, Problem.class);
-      return objectMapper.readValue(response.getBody(), Problem.class);
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
-  }
-
   protected InputStream createRequest(Map<String, String> queryParams) {
     try {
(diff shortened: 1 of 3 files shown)


torbjokv (approved):
Looks good as far as I see :-) I have not used @Nested before but it helps with readability so 👍

index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/SearchNviCandidatesHandlerTestBase.java
```diff
@@ -0,0 +1,126 @@
...
+      Map<String, String> queryParams) {
+    var request = createRequest(queryParams);
+    try {
+      handler.handleRequest(request, output, CONTEXT);
+      var response = GatewayResponse.fromOutputStream(output, PaginatedSearchResult.class);
+      output.reset();
+      return objectMapper.readValue(response.getBody(), SEARCH_RESULT_TYPE);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
```
truhacevkir, line 70: You can reuse similar method with another signature.
return handleBadRequest(createRequest(queryParams));

truhacevkir (approved):

PR #525, opened 2025-07-03
[np-49374] feat: Add filter for exclude unassigned

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49374

This adds a new query parameter to the API to filter by whether or not the user's organization has a assigned a curator to the candidate. The default value is `excludeUnassigned=false`. The query parameter `assignee=<username` now ignores unassigned candidates.

Examples:

- `?assignee=MrCurator&excludeUnassigned=true`: This returns only those candidates assigned to `MrCurator`
- `?assignee=MrCurator`: This returns candidates assigned to `MrCurator` and also any candidates without an assignee from their organization.
- `?excludeUnassigned=true`: This returns all candidates that have an assignee

### DIFF
--- a/index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/SearchNviCandidatesHandlerIntegrationTest.java
+++ b/index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/SearchNviCandidatesHandlerIntegrationTest.java
@@ -257,14 +257,11 @@ class SearchNviCandidatesHandlerIntegrationTest extends SearchNviCandidatesHandl
     }
 
     private static Map<String, NviCandidateIndexDocument> createDocumentsForAssigneeTests() {
-      var assignedToCurrentUser =
-          randomApproval(OUR_USER, OUR_ORGANIZATION, randomOrganizationId());
-      var assignedToOtherUser =
-          randomApproval(OUR_OTHER_USER, OUR_ORGANIZATION, randomOrganizationId());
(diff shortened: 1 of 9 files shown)


index-handlers/src/test/java/no/sikt/nva/nvi/index/apigateway/SearchNviCandidatesHandlerIntegrationTest.java
```diff
@@ -190,6 +210,115 @@ void shouldFilterByReportingPeriodAndNotPublicationYear() {
...
+          documentWithApprovals(UNASSIGNED_BY_THEM, assignedToOtherUser, noAssigneeFromThem),
+          UNASSIGNED_BY_ALL,
+          documentWithApprovals(UNASSIGNED_BY_ALL, noAssigneeFromUs, noAssigneeFromThem),
+          UNRELATED_ASSIGNED,
+          documentWithApprovals(UNRELATED_ASSIGNED, assignedToTheirUser),
+          UNRELATED_UNASSIGNED,
+          documentWithApprovals(UNRELATED_UNASSIGNED, noAssigneeFromThem));
+    }
+
+    private static Stream<Arguments> assigneeTestCaseProvider() {
```
LarsV123, line 283: This is admittedly not the most readable test setup, but we need to test a wide set of combinations of assignee status for the user organization and other organizations. Suggestions are welcome.

torbjokv (approved):

PR #528, opened 2025-07-04
fix: Handle missing top-level organizations during import

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49433

Fix to handle imported NVI results with no affiliated organizations. This is the same handling as we do during BatchScan.

### DIFF
--- a/event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinNviReportEventConsumer.java
+++ b/event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinNviReportEventConsumer.java
@@ -1,6 +1,8 @@
 package no.sikt.nva.nvi.events.cristin;
 
+import static java.util.Collections.emptyList;
 import static no.sikt.nva.nvi.common.db.DynamoRepository.defaultDynamoClient;
+import static no.sikt.nva.nvi.common.utils.Validator.isMissing;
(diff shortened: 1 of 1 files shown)


event-handlers/src/main/java/no/sikt/nva/nvi/events/cristin/CristinNviReportEventConsumer.java
```diff
@@ -109,6 +111,20 @@ private DbCandidate createUpdatedDbCandidate(
...
+      // imported results are missing data.
+      logger.error(
+          "Missing top level organizations for publication with identifier {}",
+          publication.identifier());
+      return emptyList();
+    }
+    return publication.topLevelOrganizations().stream()
+        .map(Organization::toDbOrganization)
+        .toList();
+  }
```
brinxmat, line 126: Typically, I'd recommend not commenting, but use the method name. Also, single return points are nice.

```suggestion
    return isMissing(publication.topLevelOrganizations()) 
      ? createEmptyListForImportedResultsThatLackData() 
      : publication.topLevelOrganizations().stream()
        .map(Organization::toDbOrganization)
        .toList();
  }

  private List<DbOrganization> createEmptyListForImportedResultsThatLackData() {
      logger.error(
          "Missing top level organizations for publication with identifier {}",
          publication.identifier());
      return emptyList();
    }
```

brinxmat (approved):

PR #529, opened 2025-07-04
[np-49370] feat: Add query parameter for global  approval status

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49370

Changes:
- feat: Add a query parameter to filter by one of multiple global approval statuses
- fix: Change the value "disputed" to "dispute" in the OpenAPI spec for convenience (matching the enum)


index-handlers/src/main/java/no/sikt/nva/nvi/index/query/CandidateQuery.java
```diff
@@ -289,7 +288,7 @@ public String getFilter() {
 
   public static class Builder {
 
-    private List<String> affiliationIdentifiers;
+    private List<String> affiliationIdentifiers = emptyList();
```
brinxmat, line 291: These are updated or replaced?

brinxmat (approved):

PR #533, opened 2025-07-07
[np-49381] refactor: Simplify file structure

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49371

This is incidental refactoring to remove linter warnings and make the structure easier to parse.

Changes:

- refactor: Convert `CandidateQuery` to a record and remove the duplicate inner class for parameters
- refactor: Extract the enum `QueryFilterType` to a separate file
- refactor: Combine relevant parameters into a single `ApprovalQuery` record in the builder instead of passing them through one by one

### DIFF
--- a/index-handlers/src/main/java/no/sikt/nva/nvi/index/query/CandidateQuery.java
+++ b/index-handlers/src/main/java/no/sikt/nva/nvi/index/query/CandidateQuery.java
@@ -2,7 +2,6 @@ package no.sikt.nva.nvi.index.query;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptySet;
-import static java.util.Objects.isNull;
 import static java.util.Objects.nonNull;
 import static no.sikt.nva.nvi.common.utils.JsonUtils.jsonPathOf;
 import static no.sikt.nva.nvi.index.model.document.ApprovalStatus.APPROVED;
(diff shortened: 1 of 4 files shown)


index-handlers/src/main/java/no/sikt/nva/nvi/index/query/QueryFilterType.java
```diff
@@ -0,0 +1,42 @@
...
+  ASSIGNMENTS_AGG("assignments"),
+  EMPTY_FILTER("");
+
+  private final String filter;
+
+  QueryFilterType(String filter) {
+    this.filter = filter;
+  }
+
+  public static Optional<QueryFilterType> parse(String candidate) {
```
brinxmat, line 31: The use of `candidate` here, while correct linguistically, conflicts with the domain term "candidate"

brinxmat (approved):

index-handlers/src/main/java/no/sikt/nva/nvi/index/query/QueryFilterType.java
LarsV123: Agreed. This is the original code though, I just moved it as-is. I can rename it now for clarity since I'm at it.

index-handlers/src/main/java/no/sikt/nva/nvi/index/query/CandidateQuery.java
```diff
@@ -56,37 +53,18 @@
...
+    ApprovalQuery approvalQuery,
+    List<String> affiliations,
+    boolean excludeSubUnits,
+    QueryFilterType filter,
+    String username,
+    String topLevelCristinOrg,
+    String searchTerm,
+    String year,
+    String category,
+    String title,
```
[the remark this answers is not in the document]
LarsV123: For most of them, not really. We could do regex-validation on the `year`, and the `topLevelCristinOrg` should be an URI (so we could map it back and forth), but they are strings both when coming in (from query parameters) and going out (as OpenSearch queries).

PR #544, opened 2025-08-28
Use nt in expanded resources

brinxmat:
Motivation: 

Basically, since we are processing a graph, provide a verbose, portable graph representation of the data to save on processing the complex, human-readable JSON data. Processing JSON-LD has a multi-second overhead above processing N-Triples, and this will save on processing any data in NVI. 

I have implemented code that allows processing of JSON-LD _and_ N-Triples, so this is backward compatible with the current status quo in expanded resources and will seamlessly apply when the "nt" node is available in an expanded resource.

Pros:
- No JSON-LD context-swapping trickery
- Better data processing performance
- Closer representation to the data we are actually using

Cons:
- It isn't JSON, which is maybe an expectation
- Currently not implemented in nva-publication-api


nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/NvaGraph.java
```diff
@@ -32,14 +38,22 @@ public NviGraph toNviGraph() {
...
     try {
       RDFDataMgr.read(model, inputStream, Lang.JSONLD);
     } catch (RiotException e) {
       logInvalidJsonLdInput(e);
     }
   }
 
+  private static void loadNtriplesIntoModel(Model model, InputStream inputStream) {
+    try {
+      RDFDataMgr.read(model, inputStream, Lang.NTRIPLES);
```
brinxmat, line 51: Same thing, different serialization -> time savings.

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/NvaGraph.java
```diff
@@ -19,7 +19,13 @@ public record NvaGraph(Model model) implements GraphValidable {
 
   public static NvaGraph fromJsonLd(JsonNode content) {
     var model = ModelFactory.createDefaultModel();
-    loadDataIntoModel(model, stringToStream(content.toString()));
+    loadJsonLdIntoModel(model, stringToStream(content.toString()));
```
brinxmat, line 22: A bit sad for some data to be wrapped in JSON, but hey "attributes" and "backward compatibility"

LarsV123 (approved):
Nice! Depends on fixes in downstream branches, but otherwise looks good.

PR #545, opened 2025-09-01
Fix tests for nt transformation

brinxmat:
- Fixes mappings of some things in contexts and JSON-POJOs
  - Are my assumptions correct? 
  - Note that the Cristin mappings are redundant once we are in publication-api-land (i.e. expansion _should_ re-map values from Org-API to _true_ NVA-semantics)
- Still couple of failing tests due to ordering following batch scan migrations…maybe you can fix easily as you are maybe familiar with the code?


nvi-commons/src/test/java/no/sikt/nva/nvi/common/service/PublicationLoaderServiceTest.java
```diff
@@ -1,17 +1,61 @@
 package no.sikt.nva.nvi.common.service;
 
+import static no.sikt.nva.nvi.common.examples.ExamplePublications.CONTRIBUTOR_NO_AFFILIATION;
```
brinxmat, line 3: The changes here were due to my "fixing" the import of one test and then reformatting…there is no real change here.

nvi-commons/src/testFixtures/resources/expandedPublications/validation/organizationRepeatedCountry.json
```diff
@@ -165,7 +165,7 @@
         "name" : "Ola Nordmann",
         "verificationStatus" : "Verified"
       } ],
-      "countryCode" : "NO",
+      "countryCode" : [ "NO", "GB" ],
```
brinxmat, line 168: This was added as I could not see the original repeat

nvi-test/src/main/resources/template_publication.json
```diff
@@ -70,6 +69,9 @@
       "@container" : "@set",
       "@id" : "contributor"
     },
+    "countryCode": {
```
brinxmat, line 72: It is always "country" in RDF.

nvi-commons/src/main/java/no/sikt/nva/nvi/common/client/model/Organization.java
```diff
@@ -20,7 +20,7 @@
 
 public record Organization(
     @JsonProperty("id") URI id,
-    @JsonProperty("country") String countryCode,
+    @JsonProperty("countryCode") String countryCode,
```
brinxmat, line 23: In JSON, it is "countryCode"

brinxmat (commented):
On the whole, I think the errors came from relying on JSON and the mappings working in one direction, but not the inverse. Probably best to move everything to pub-api to avoid having to repeat and copy contexts, frames etc.

LarsV123 (approved):
See comment on `@Context` node and separate PR for unrelated failing tests, otherwise this looks correct!

PR #550, opened 2025-09-09
feat: Add logging to upsert events

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49728

Changes: 

- Adds detailed logging around events that create/update significant data (candidates, approvals, reporting periods) to make future debugging easier. Note that this includes the username and organization ID where appropriate.
- Replaces a few calls to `.collect(Collectors.toList())` with `toList()`


nvi-commons/src/test/java/no/sikt/nva/nvi/common/db/CandidateRepositoryTest.java
LarsV123, line 1: This is just automatic reformatting, no changes here

truhacevkir (approved):

PR #749, opened 2026-04-15
refactor: Extricate environment from Candidate record

LarsV123:
Removes `Environment` field from `Candidate` because this does not belong there and few callers use it. The only use-case was to provide environment-specific URLs for Candidate ID and the generic Context endpoint. This is now provided by a new `EnvironmentUriFactory` instead.


nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/Candidate.java
```diff
@@ -70,14 +68,9 @@ public record Candidate(
     ReportStatus reportStatus,
     Instant reportedDate,
     Long revision,
-    UUID version,
-    // TODO: Remove environment from this record
-    Environment environment) {
+    UUID version) {
```
LarsV123, line 76: This is the main change: `Candidate` no longer contains an `Environment` field

nvi-commons/src/main/java/no/sikt/nva/nvi/common/utils/EnvironmentUriFactory.java
```diff
@@ -0,0 +1,32 @@
...
+  private EnvironmentUriFactory() {}
+
+  public static URI candidateId(Environment environment, UUID candidateIdentifier) {
+    return baseUrl(environment)
+        .addChild(CANDIDATE_PATH)
+        .addChild(candidateIdentifier.toString())
+        .getUri();
+  }
+
+  public static URI context(Environment environment) {
```
torbjokv, line 24: candidateContext maybe? in case we get more than one?

torbjokv (approved):

nvi-commons/src/main/java/no/sikt/nva/nvi/common/utils/EnvironmentUriFactory.java
LarsV123: Maybe, but lets rename it when that happens. The URI in question is `https://api.nva.unit.no/scientific-index/context` etc, so already "globally scoped".

PR #751, opened 2026-04-15
[np-51079] feat Add reported date

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-51079

Changes:

- Adds `reportedDate` field to candidate models (domain, database, and index)
- Adds methods to `Candidate` and `CandidateService` to update a candidate to "reported" so that it becomes immutable
- Adds helper method `TestScenario.setupReportedCandidate` to simplify setting up test cases

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1329,6 +1329,15 @@ components:
               type: string
             startDate:
               type: string
+        status:
+          type: string
+          description: The report status of the candidate
+          example: "Reported"
+        reportedDate:
+          type: string
+          format: date-time
+          description: Timestamp for when the candidate was marked as reported
+          example: "2026-01-15T12:00:00Z"
         allowedOperations:
           type: array
           description: >
--- a/integration-tests/src/test/java/cucumber/steps/EvaluationSteps.java
+++ b/integration-tests/src/test/java/cucumber/steps/EvaluationSteps.java
@@ -240,14 +240,6 @@ public class EvaluationSteps {
     for (var institutionId : candidate.approvals().keySet()) {
       scenario.updateApprovalStatus(candidate.identifier(), ApprovalStatus.APPROVED, institutionId);
     }
-    //    candidate
-    //        .approvals()
-    //        .keySet()
-    //        .forEach(
-    //            institutionId ->
-    //                scenario.updateApprovalStatus(
-    //                    candidate.identifier(), ApprovalStatus.APPROVED, institutionId));
-    //    var refreshedCandidate = scenario.getCandidateByIdentifier(candidate.identifier());
     scenario.getCandidateService().reportCandidate(candidate.identifier(), Instant.now());
   }
 }
--- a/integration-tests/src/test/java/cucumber/steps/EvaluationSteps.java
+++ b/integration-tests/src/test/java/cucumber/steps/EvaluationSteps.java
@@ -237,14 +237,17 @@ public class EvaluationSteps {
   }
 
   private void setCandidateToReported(Candidate candidate) {
-    candidate
-        .approvals()
-        .keySet()
-        .forEach(
-            institutionId ->
-                scenario.updateApprovalStatus(
-                    candidate.identifier(), ApprovalStatus.APPROVED, institutionId));
-    var refreshedCandidate = scenario.getCandidateByIdentifier(candidate.identifier());
-    scenario.getCandidateService().reportCandidate(refreshedCandidate.identifier(), Instant.now());
+    for (var institutionId : candidate.approvals().keySet()) {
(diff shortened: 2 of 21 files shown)


integration-tests/src/test/java/cucumber/steps/EvaluationSteps.java
```diff
@@ -172,7 +170,7 @@ private void setupApplicablePublication(PublicationDate publicationDate) {
         publicationBuilder
             .withPublicationDate(publicationDate)
             .withContributor(verifiedCreatorFrom(nviOrganization))
-            .withContributor(unverifiedCreatorFrom(nviOrganization));
```
LarsV123, line 175: This changes default test data so that the default case is a "reportable" candidate. Test cases that need an unverified creator should be explicitly adding one.

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/CandidateService.java
```diff
@@ -143,6 +144,18 @@ public void updateCandidate(UpsertNonNviCandidateRequest request) {
...
+  /**
+   * Marks a Candidate as `reported`, which means it is included in the final report for the
+   * corresponding reporting period and is now immutable.
+   */
+  public Candidate reportCandidate(UUID candidateIdentifier, Instant reportedDate) {
+    LOGGER.info("Updating candidate with identifier={} to reported", candidateIdentifier);
+    var candidate = getCandidateByIdentifier(candidateIdentifier);
+    var reportedCandidate = candidate.updateToReportedCandidate(reportedDate);
+    updateCandidate(reportedCandidate);
+    return reportedCandidate;
```
torbjokv: This returns the in-memory reportedCandidate rather than re-fetching from the database. The revision field (used for optimistic locking) on the returned object will be stale. Probably not returning is a better idea?

docs/openapi.yaml
```diff
@@ -1329,6 +1329,15 @@ components:
               type: string
             startDate:
               type: string
+        status:
```
torbjokv, line 1332: is this a candidate status or a report status. should it be renamed to reportStatus? (maybe that breaks stuff, I don't know, but I just thows it out there)

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/Candidate.java
```diff
@@ -174,6 +176,20 @@ public Candidate updateToNonCandidate() {
         .build();
   }
 
+  public Candidate updateToReportedCandidate(Instant reportedDate) {
+    if (isReported()) {
+      throw new IllegalCandidateUpdateException(CANDIDATE_IS_REPORTED);
+    }
+    if (getGlobalApprovalStatus() != GlobalApprovalStatus.APPROVED) {
+      throw new IllegalCandidateUpdateException("Cannot report non-approved candidate");
```
torbjokv, line 184: nitpick: string constant

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/Candidate.java
```diff
@@ -174,6 +176,20 @@ public Candidate updateToNonCandidate() {
         .build();
   }
 
+  public Candidate updateToReportedCandidate(Instant reportedDate) {
+    if (isReported()) {
```
torbjokv, line 180: should we also check if period is in the correct state here?

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/Candidate.java
LarsV123: ErrorProne actually flags that as an issue and recommends inlining single-use String constants (e.g. https://errorprone.info/bugpattern/InlineFormatString).

docs/openapi.yaml
LarsV123: It's report status and it should probably be renamed, but not now. It's an existing field that we're just adding to the docs, so that would probably break something in the frontend.

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/CandidateService.java
LarsV123: Yes, good point!

torbjokv (approved):

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/Candidate.java
LarsV123: Made testing a bit awkward, but good point. Fixed now.

PR #761, opened 2026-05-05
refactor: Split nvi-commons into smaller modules

LarsV123:
Splits `nvi-commons` by extracting related functionality with significant dependencies into smaller modules. This will enable us to limit the dependencies pulled in by downstream handler modules (left for a future PR).

Changes:

- Creates four new Gradle modules under the `libs/` folder for functionality extracted from `nvi-commons`
- Adds explicit imports for transitive dependencies we relied on (´jakarta-json-api`, `parsson`, `titanium-json-ld`)
- Excludes deprecated `org.glassfish:jakarta.json` artifact that overlaps with the API from other dependencies


libs/viewing-scope/src/main/java/no/sikt/nva/nvi/viewingscope/utils/GraphUtils.java
```diff
@@ -1,4 +1,4 @@
-package no.sikt.nva.nvi.common.utils;
+package no.sikt.nva.nvi.viewingscope.utils;
```
torbjokv: Does GraphUtils belong to viewingscope?

libs/migration-service/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+dependencies {
+    api(nvaLibs.clients)
+    api(project(':nvi-commons'))
+    implementation(nvaCatalog.slf4j.api)
+    implementation(nvaLibs.core)
+    implementation(project(':publication-service'))
+    runtimeOnly(nvaCatalog.aws.apache.client)
+    runtimeOnly(nvaCatalog.aws.sdk2.urlconnection)
+    runtimeOnly(nvaCatalog.jackson.databind)
+    runtimeOnly(nvaCatalog.jena.arq)
```
torbjokv: do we need jena stuff here? probably comes through publication-service

torbjokv (approved):

libs/migration-service/build.gradle
LarsV123: Nope, good catch. Added because `buildHealth` tends to say "Do X,Y,. and Z" in the generated report, but once you've done X and Y, there's no need to do Z...

libs/viewing-scope/src/main/java/no/sikt/nva/nvi/viewingscope/utils/GraphUtils.java
LarsV123: No, not really. I'll split it to a separate module.

PR #775, opened 2026-06-10
refactor: Improve SPARQL queries

LarsV123:
Changes:

- [fix: Evaluate only once per publication in isInternationalCollaboration check](https://github.com/BIBSYSDEV/nva-nvi/pull/775/changes/9a3cdddae5203fd8a731e474e0c1c28065ef9be4)
- [fix: Remove unnecessary optional wrapper with unstable semantics](https://github.com/BIBSYSDEV/nva-nvi/pull/775/changes/14e442b83839964146f25e8b46b99bf230a43b6a)
- Splits the query `nvi_projection.rq` into three smaller queries
- Removes irrelevant publication channels from query output, so that a `PublicationChannelDto` will only have a single valid channel and `PointService` no longer needs to determine which one to use


nvi-commons/src/main/java/no/sikt/nva/nvi/common/dto/PublicationDto.java
```diff
@@ -99,6 +99,10 @@ public static PublicationDto from(String json) throws JsonProcessingException {
     return dtoObjectMapper.readValue(json, PublicationDto.class);
   }
 
+  public PublicationChannelDto getNviChannel() {
+    return publicationChannels.stream().filter(PublicationChannelDto::isValid).toList().getFirst();
+  }
+
```
LarsV123, line 105: Temporary method to limit changes. The field can be converted from `Collection<PublicationChannelDto>` to just `PublicationChannelDto` in a separate PR.

brinxmat (approved):

PR #776, opened 2026-06-15
fix: Expose NviPeriod status in all DTOs

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-51058

Exposes the current status of an NVI period in all end-points that return data about an NVI period.

Changes:

- Extends the exposed data model to include status for all NVI periods.
- Merges two overlapping DTOs used to represent NVI periods in different endpoints.


docs/openapi.yaml
```diff
@@ -1113,7 +1113,10 @@ paths:
                     period:
                       type: NviPeriod
                       id: https://api.nva.unit.no/scientific-index/period/2025
-                      year: "2025"
```
LarsV123, line 1116: Note the change here: `year -> publishingYear`. This was done to match the field frontend was actually reading from the now-deleted `PeriodStatusDto`.

torbjokv (approved):

PR #778, opened 2026-06-22
fix(evaluator): preserve closed-but-not-reported candidates from stripping

torbjokv:
Prevents an approved but not-yet-reported candidate in a closed period from having its approvals changed by a publication edit, so closed-period numbers can never change. Previously the non-candidate decision ran before the closed-period guard, so a publication edit could strip the candidate in place; a year correction could also move it out of the closed period.

- Reorder `EvaluatorService.evaluateCandidacy` so the closed-period guard for an existing candidate runs before the non-candidate and move decisions
- Freeze any existing candidate in a closed period: it is neither stripped in place nor moved out, regardless of how the publication is edited or which year it gets. Correcting the year for such a candidate requires reopening the period
- Add a backstop period guard in `Candidate.updateToNonCandidate()` so the invariant holds even if a caller bypasses the evaluator
- Update `Evaluation.feature`: closed-period candidates are not demoted and not moved; demotion/move apply only in pending or open periods
- Add reproducing tests for the closed-but-not-reported window and the year-correction case

Note: NP-51363 originally allowed moving a closed-period candidate to a new open period. This was intentionally tightened to a full freeze, since such a move also changes the closed period's numbers.

https://sikt.atlassian.net/browse/NP-51363

Changes previous implementation done here https://sikt.atlassian.net/browse/NP-48355

### DIFF
--- a/event-handlers/src/main/java/no/sikt/nva/nvi/events/evaluator/EvaluatorService.java
+++ b/event-handlers/src/main/java/no/sikt/nva/nvi/events/evaluator/EvaluatorService.java
@@ -174,25 +174,26 @@ public class EvaluatorService {
     if (optionalCandidate.isEmpty() || !isInClosedPeriod(optionalCandidate.get())) {
       return false;
     }
-    if (movesToOpenPeriod(candidateAndPeriods, publicationDate)) {
-      logger.info(
-          "Candidate {} moves from a closed period to open period {}",
-          optionalCandidate.get().identifier(),
-          publicationDate.year());
-      return false;
+    var candidate = optionalCandidate.get();
+    if (remainsInSamePeriod(candidate, publicationDate)) {
+      return true;
     }
-    return true;
+    logger.info(
+        "Candidate {} in a closed period is being moved to year {}",
+        candidate.identifier(),
+        publicationDate.year());
+    return false;
   }
 
   private static boolean isInClosedPeriod(Candidate candidate) {
     return candidate.getPeriod().map(NviPeriod::isClosed).orElse(false);
   }
 
-  private boolean movesToOpenPeriod(
-      CandidateAndPeriods candidateAndPeriods, PublicationDateDto publicationDate) {
-    return candidateAndPeriods
-        .getPeriod(publicationDate.year())
-        .map(NviPeriod::isOpen)
+  private static boolean remainsInSamePeriod(
+      Candidate candidate, PublicationDateDto publicationDate) {
+    return candidate
+        .getPeriod()
+        .map(period -> period.hasPublishingYear(publicationDate.year()))
         .orElse(false);
   }
 
--- a/event-handlers/src/main/java/no/sikt/nva/nvi/events/evaluator/EvaluatorService.java
+++ b/event-handlers/src/main/java/no/sikt/nva/nvi/events/evaluator/EvaluatorService.java
@@ -97,6 +97,14 @@ public class EvaluatorService {
       return Optional.empty();
     }
 
+    // Preserve an existing candidate whose period is closed before any non-candidate decision
+    // below, so a publication edit cannot strip a closed-but-not-yet-reported candidate.
+    if (shouldPreserveExistingCandidateInClosedPeriod(
+        candidateAndPeriods, publication.publicationDate())) {
+      logger.info(SKIPPED_EVALUATION_MESSAGE, publication.id());
+      return Optional.empty();
+    }
+
     // Check that the publication meets the basic requirements to be a candidate
     if (isNonCandidate(publication)) {
       return createNonNviCandidateMessage(publication.id());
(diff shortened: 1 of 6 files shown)


event-handlers/src/test/java/no/sikt/nva/nvi/events/evaluator/EvaluateNviCandidateHandlerTest.java
```diff
@@ -669,21 +669,50 @@ void shouldNotThrowErrorWhenEvaluatingExistingCandidateInClosedPeriod() {
...
+      assertThat(candidateAfterUpdate.isApplicable()).isTrue();
+      assertThat(candidateAfterUpdate.approvals()).isEqualTo(candidateBeforeUpdate.approvals());
+      assertThat(candidateAfterUpdate.revision()).isEqualTo(candidateBeforeUpdate.revision());
+    }
+
+    @Test
+    void shouldMoveExistingClosedPeriodCandidateWhenYearCorrectedToOpenPeriod() {
+      // Given an applicable Candidate in a closed period
+      // When the publication year is corrected to a different, open period
+      // Then the Candidate must be moved and re-evaluated, not frozen
```
joachimjorgensen: I do not believe this should be allowed. If the period is closed, then no candidates should be altered. Maybe check with PO.

event-handlers/src/main/java/no/sikt/nva/nvi/events/evaluator/EvaluatorService.java
```diff
@@ -166,16 +168,32 @@ private boolean isPublished(PublicationDto publication) {
     return nonNull(publication.status()) && "published".equalsIgnoreCase(publication.status());
   }
 
-  private boolean isApplicableCandidateInClosedPeriod(
+  private boolean shouldPreserveExistingCandidateInClosedPeriod(
```
joachimjorgensen: I think this method is not needed. It will only guard against changes done to a candidate within i period. While correct, it would maybe be better to guard the entire closed period.

event-handlers/src/main/java/no/sikt/nva/nvi/events/evaluator/EvaluatorService.java
joachimjorgensen, line 135:
```suggestion
    }
    
    if (/* Candidate is in closed period */) { return true; }
```

joachimjorgensen: Something like this enough?

event-handlers/src/test/java/no/sikt/nva/nvi/events/evaluator/EvaluateNviCandidateHandlerTest.java
torbjokv: Agreed, and that is the decision we landed on after checking with the PO. A candidate in a closed period is now frozen entirely — neither stripped in place nor moved out. This test was renamed to `shouldPreserveExistingCandidateInClosedPeriodWhenYearCorrectedToOpenPeriod` and now asserts the candidate is preserved (unchanged, stays in its period).

event-handlers/src/main/java/no/sikt/nva/nvi/events/evaluator/EvaluatorService.java
torbjokv: Done — the guard now protects the entire closed period rather than just same-period changes. Simplified to a single `isExistingCandidateInClosedPeriod(...)` check and removed the per-year/move logic.

event-handlers/src/main/java/no/sikt/nva/nvi/events/evaluator/EvaluatorService.java
torbjokv: Good call — folded it into `shouldSkipEvaluation` so all skip reasons live in one place (invalid year, reported, closed period). See latest commit.

event-handlers/src/main/java/no/sikt/nva/nvi/events/evaluator/EvaluatorService.java
```diff
@@ -138,6 +133,13 @@ private boolean shouldSkipEvaluation(
       logger.warn(REPORTED_CANDIDATE_MESSAGE);
       return true;
     }
+
+    // Freeze any existing candidate in a closed period: closed-period numbers must never change,
+    // so it is neither stripped in place nor moved out by a publication edit.
+    if (isExistingCandidateInClosedPeriod(candidateAndPeriods)) {
```
joachimjorgensen, line 139: To syncronize with the isReported-check above, either inline the method in the if or extract the above isReported check into a similar method call?

joachimjorgensen (approved):
Liten kommentar, men ellers bra!

brinxmat (approved):

event-handlers/src/main/java/no/sikt/nva/nvi/events/evaluator/EvaluatorService.java
torbjokv: extracted isReportedCandidate 👍

PR #781, opened 2026-07-01
[np-51434] test: Add tests to prevent regressions

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-51434

This PR is split from a larger change to make it easier to review and less likely that the larger change breaks something.

Goals:

1. Add tests covering existing behavior to prevent regressions
2. Add disabled tests covering the bug in question
3. Add placeholder scenarios for related bugs that were uncovered while investigating this bug

Changes:

- [refactor: Rename Cucumber tag to match JUnit equivalent](https://github.com/BIBSYSDEV/nva-nvi/pull/781/commits/d145d0a1b721d701b6aff6059e6757028c1425f1)
- [refactor: Improve ContributorFixtures](https://github.com/BIBSYSDEV/nva-nvi/pull/781/commits/b4e61c9b164793f2bc96a265da120319e43ac7b4)
- [test: Add test fixtures to fake Cristin org responses](https://github.com/BIBSYSDEV/nva-nvi/pull/781/commits/4726aece4ee31bfbad91a28c626c348a36ffa05c)
- [test: Add tests and test placeholders for index feature](https://github.com/BIBSYSDEV/nva-nvi/pull/781/commits/37d04607a194793db616e752303ced21f53f4fe3)
- [chore: Remove unused test constants](https://github.com/BIBSYSDEV/nva-nvi/pull/781/commits/c69b67cdacd3853ec89f1920cdfd70ccaaba96a0)


integration-tests/src/test/resources/features/indexing/IndexingReportedCandidates.feature
LarsV123, line 1: This is the file that is most relevant to review in this PR, everything else is plumbing.

integration-tests/src/test/resources/features/indexing/IndexingReportedCandidates.feature
```diff
@@ -0,0 +1,95 @@
...
+      When the Candidate is indexed
+      Then the indexed NVI points match the Candidate
+      And the indexed NVI creators match the Candidate
+
+    @Disabled # FIXME: NP-51406 - Reported NVI numbers can silently change when a candidate is reindexed
+    Scenario: Creator affiliation differs between Candidate and Publication, index uses the Candidate
+      Given the creator in section A1 is moved to section A2 in the Publication
+      When the Candidate is indexed
+      Then the indexed NVI points match the Candidate
+      And the indexed NVI affiliations match the Candidate
```
LarsV123: Failing tests that describe the relevant bug

integration-tests/src/test/java/cucumber/contexts/IndexingContext.java
```diff
@@ -0,0 +1,102 @@
...
+
+  private final TestScenario scenario;
+
+  // Stubs the Cristin organization registry. Unlike evaluation (which reads the org hierarchy from
+  // the expanded publication), the index document generator looks up each NVI affiliation's
+  // hierarchy live against the Cristin API (OrganizationRetriever/UriRetriever), ignoring the
+  // partOf already present in the publication. So indexing tests must register a canned org
+  // response per affiliation instead of relying on the input document. This live-lookup fragility
+  // is the tech debt behind NP-51406 and NP-51432; once indexing derives org data from persisted
+  // data, this stub becomes unnecessary.
```
brinxmat, line 45: Multiline comment

integration-tests/src/test/java/cucumber/steps/IndexingSteps.java
```diff
@@ -0,0 +1,359 @@
...
+    var publicationDetails = candidate.publicationDetails();
+    return Stream.concat(
+            publicationDetails.verifiedCreators().stream()
+                .map(verifiedCreator -> verifiedCreator.id().toString()),
+            publicationDetails.unverifiedCreators().stream().map(UnverifiedNviCreatorDto::name))
+        .collect(Collectors.toSet());
+  }
+
+  // Compares by value string, which holds for decided approvals; a reported Candidate's approvals
+  // are all decided, so the index applies no NEW (pending-and-unassigned) transform here.
```
brinxmat: Comment-style

integration-tests/src/test/resources/features/indexing/IndexingReportedCandidates.feature
```diff
@@ -0,0 +1,95 @@
+Feature: Indexing of reported NVI Candidates
+  # Terminology
+  #
+  # NVI data: Reportable data that affects NVI points (creators that qualify, publication channel
+  #   and level used for points, points per institution, approval status). Creator name is also NVI
+  #   data and is meant to be frozen on the Candidate (see name note below).
+  # Period: NVI reporting period, currently one calendar year, with a state (pending/open/closed).
```
brinxmat, line 7: This is the crux of the issue we uncovered during refinement, there are two things:

1 Reporting period
2 Administrative reporting period

Where (1) is currently equivalent to a calendar year for _one_ reporting scope — and may be of varying lengths based on changing requirements, and (2) is the time-frame for actually doing the reporting work, and is currently from June-ish to the end of April-ish.

The Reporting period should "close", but administrative reporting should continue until it is done, I suspect.

integration-tests/src/test/resources/features/indexing/IndexingReportedCandidates.feature
```diff
@@ -0,0 +1,95 @@
...
+      And the indexed NVI affiliations match the Candidate
+      And the indexed NVI creators match the Candidate
+      And the indexed reporting status matches the Candidate
+      And the indexed approval statuses match the Candidate
+
+    Scenario: Creator added to the Publication is not indexed as an NVI creator
+      Given a creator is added to the Publication
+      When the Candidate is indexed
+      Then the added creator is not indexed as an NVI creator
+      And the indexed NVI affiliations match the Candidate
```
brinxmat: I understand the desire for brevity, but I think it is important to ensure that we express exact expectations, because this sentence is load-bearing in this and every other case in this file.

```suggestion
      And the indexed NVI affiliations are exactly the Candidate's NVI affiliations
```

nvi-commons/src/testFixtures/java/no/sikt/nva/nvi/common/cristin/CristinOrganizationFixtures.java
```diff
@@ -0,0 +1,73 @@
...
+public final class CristinOrganizationFixtures {
+
+  private static final String ORGANIZATION_CONTEXT =
+      "https://bibsysdev.github.io/src/organization-context.json";
+
+  private CristinOrganizationFixtures() {}
+
+  public static FakeCristinOrganization.Builder randomCristinOrganization(URI organizationId) {
+    var unitName = randomUnitName();
+    var labels =
```
brinxmat, line 28: Does anything happen if this list is expanded, contains blanks, nulls, etc.?

nvi-test/src/main/java/no/sikt/nva/nvi/test/TestUtils.java
```diff
@@ -62,10 +63,29 @@ public static String randomYear() {
...
+
   public static String randomTitle() {
     return String.format("%s %d", FAKER.book().title(), randomInteger(CURRENT_YEAR));
   }
 
+  public static String randomUnitName() {
+    return String.format("%s %s", FAKER.word().adjective(), FAKER.word().noun());
+  }
+
+  public static String randomEnglishUnitLabel(String unitName) {
```
brinxmat, line 78: This cannot be fixed now.

There is an issue with the name "Unit", given that it inherits a lot of Cristin's conceptualisation, from which Cristin itself has strayed. We should be careful to name things very precisely with precise semantics. 

If we assume Cristin's four-level DAG, we are ignoring the fact that Cristin no longer specifies a four-level structure, or, it turns out a DAG.

We will probably end up having to consume this at some point, and we will wonder why things don't work…and we have encoded various unmet expectations everywhere in the code.

brinxmat (approved):
I added comments, there are a few details and some fundamental issues that cannot be addressed in this PR; unfortunately, we have inherited a landscape that we now have to deal with :D

nvi-test/src/main/java/no/sikt/nva/nvi/test/TestUtils.java
LarsV123: True enough, but these are just convenience methods to get readable values for tests. It's pretty hard to debug when every single value is `randomString()`. The point was just to distinguish names between organizations, people, channels etc, and to distinguish English/Norwegian example values from each other.

I'll probably move a lot of this over to `nva-commons` (we need the same thing in several repositories), so we can rename it then if you like.

integration-tests/src/test/java/cucumber/contexts/IndexingContext.java
LarsV123: Deleted in follow-up PR, so leaving it here.

integration-tests/src/test/java/cucumber/steps/IndexingSteps.java
LarsV123: Fixed in the follow-up PR

nvi-commons/src/testFixtures/java/no/sikt/nva/nvi/common/cristin/CristinOrganizationFixtures.java
LarsV123: Probably not, but in any case I managed to get rid of live look-ups against the proxy API in the follow-up PR, so this class can be deleted. Might move it to `nva-commons` and add some tests there though.

integration-tests/src/test/resources/features/indexing/IndexingReportedCandidates.feature
LarsV123: Did a larger pass to address this and made the expectations more specific.

integration-tests/src/test/resources/features/indexing/IndexingReportedCandidates.feature
LarsV123: Leaving this for now, but we should probably create a single ontology / terminology file somewhere that properly explains this. We have scattered half-explanations all over the place now.

PR #785, opened 2026-07-02
[np-51435] feat: Replace index document generator

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-51435

Replaces the existing code for generating index documents with new mappers that use the same parser as evaluation, which produces a `PublicationDto` from the expanded publication in S3.

## Goal

Replace the legacy JSON-node index-document generator (which did live Cristin organization lookups) with a generator that builds the index document purely from the persisted `Candidate` plus a `PublicationDto`, and remove the old generator completely.

When generating index documents we should:

1. Use the persisted candidate as the source of truth so that indexed NVI facts always match what we have in the database
2. Supplement this with other data sources only for fields that do not affect NVI points
3. Be tolerant of missing data. Degraded or missing data from supplemental sources should not prevent us from producing the index document.

This is the foundation for fixing multiple different indexing bugs, such as those that the regression tests in PR #781 document.

## Problems with the old implementation

- NVI contributors were derived from the publication and enriched with data from the candidate. If a contributor was removed from the publication, their points also disappeared from the index and derived reports.
- Organizations were resolved individually (via the Cristin proxy service). Slow and error-prone, with failures going to DLQ. Particularly problematic if the organization was deleted and couldn't be resolved.
- The expanded publication was parsed using `JsonNode` paths, making it tightly coupled to the document structure and difficult to work with.

## Implementation choices worth noting

- Changed the `contributors` field in the index document to be exactly the contributors from the publication, while `nviContributors` are exactly those persisted in the candidate (see code and comment).
- This implementation uses the expanded publication in S3 (`PublicationDto`) as the only external data source, because the organization metadata there is adequate and up-to-date. We could add other data sources later if needed, such as looking up publication channels by ID to get the current name.
  - No fallback for migrated fields: These fields are assumed to be present and no fallback is used. This includes `title`, `language`, `abstractText`, `pageCount`, and `publicationDate`.
  - Fault tolerance: Non-essential fields missing from both the `Candidate` and `Publication` do not throw exceptions, but the error is logged and the field omitted from the index document.
  - DLQ: If we cannot access the publication (e.g. deleted from S3) at all, then this is still treated as an exception and goes to DLQ. We may reconsider this later if it becomes an issue.

## Future work / follow-up

- We need to run a full re-indexing job in `prod` to verify this code properly handles all existing data.
- Several follow-up tickets are created and referenced in the code for minor problems with existing data, such as missing ORCID and role for persisted contributors.


index-handlers/src/main/java/no/sikt/nva/nvi/index/model/document/PublicationDetails.java
```diff
@@ -45,7 +45,7 @@ public static final class Builder {
     private String abstractText;
     private PublicationDateDto publicationDate;
     private List<NviContributor> nviContributors;
-    private List<ContributorType> contributors;
+    private List<Contributor> contributors;
```
LarsV123, line 48: Before:

- `contributors` contained all contributors, derived from the same sources and discriminated by type
- Those awarded NVI points were duplicated to the `nviContributors` field

Now:
- `contributors` contains exactly those contributors present in the expanded publication
- `nviContributors` contains exactly those contributors with points present in the `Candidate`

This was done to make the code simpler, but has the added benefit of letting us search for a renamed contributor by both the old name (persisted in DB) and the new name (if updated in the publication).

index-handlers/src/main/java/no/sikt/nva/nvi/index/mapper/ContributorMapper.java
```diff
@@ -0,0 +1,182 @@
...
+
+  private NviContributor buildNviContributor(NviCreator creator) {
+    var enrichment = findEnrichmentContributor(creator).orElse(null);
+    return NviContributor.builder()
+        .withId(creator.id())
+        .withName(extractName(creator, enrichment))
+        .withOrcid(extractOrcid(enrichment))
+        .withRole(extractRole(enrichment))
+        .withAffiliations(buildNviCreatorAffiliations(creator))
+        .build();
```
brinxmat, line 71: Possibly something like (bearing in mind that the call chain is problematic since we should be able to expose the nullity of the enrichment as Optional in the API):

```suggestion
    var nviContributor = NviContributor.builder()
        .withId(creator.id())
        .withAffiliations(buildNviCreatorAffiliations(creator));
    if (nonNull(enrichment)) {
         nviContributor.withName(extractName(creator, enrichment))
          .withOrcid(extractOrcid(enrichment))
          .withRole(extractRole(enrichment))
    }
    return nviContributor.build();
```

index-handlers/src/main/java/no/sikt/nva/nvi/index/mapper/ContributorMapper.java
```diff
@@ -0,0 +1,182 @@
...
+    return publicationContributors.stream()
+        .filter(ContributorDto::isCreator)
+        .filter(contributorDto -> matches(creator, contributorDto))
+        .findFirst();
+  }
+
+  private static boolean matches(NviCreator creator, ContributorDto contributorDto) {
+    return creator.isVerified()
+        ? creator.id().equals(contributorDto.id())
+        : nonNull(contributorDto.name()) && contributorDto.name().equals(creator.name());
```
brinxmat: DANGER WILL ROBINSON

index-handlers/src/main/java/no/sikt/nva/nvi/index/mapper/PublicationDetailsMapper.java
```diff
@@ -0,0 +1,104 @@
...
+
+  private static Pages buildPages(PageCount pageCount) {
+    if (isNull(pageCount)) {
+      return null;
+    }
+    return Pages.builder()
+        .withBegin(pageCount.first())
+        .withEnd(pageCount.last())
+        .withNumberOfPages(pageCount.total())
+        .build();
```
brinxmat: Can't help but feel that all of these mapper-driven builder methods are object methods `Pages::from`

index-handlers/src/main/java/no/sikt/nva/nvi/index/mapper/PublicationDetailsMapper.java
```diff
@@ -0,0 +1,104 @@
...
+   * whose channel ID no longer matches the Publication (channel merged or superseded)
+   */
+  private Optional<PublicationChannelDto> findMatchingPublicationChannelDto(
+      no.sikt.nva.nvi.common.model.PublicationChannel channel) {
+    return findChannelDtoById(channel.id()).or(() -> findChannelDtoByType(channel.channelType()));
+  }
+
+  private Optional<PublicationChannelDto> findChannelDtoById(URI id) {
+    return nonNull(id)
+        ? publicationChannels().stream().filter(dto -> id.equals(dto.id())).findFirst()
```
brinxmat: The semantics here and on L97, do we explicitly want the deterministic "first", or would "any" do?

index-handlers/src/test/java/no/sikt/nva/nvi/index/IndexDocumentContentTest.java
```diff
@@ -0,0 +1,501 @@
...
+  @Test
+  void shouldNotPopulateSectorInApprovalViewWhenSectorIsNull() {
+    var institutionId = randomOrganizationId();
+    var candidate = setupCandidateWithInstitutionPoints(institutionId, null, false);
+    stubPublication(candidate);
+
+    var document = generateIndexDocument(candidate);
+
+    assertThat(approvalFor(document, institutionId).sector()).isNull();
+  }
```
brinxmat: Perhaps a parameterised test?
```suggestion
  @ParameterizedTest
  @NullSource // not sure if it is possible to use @NullAndEmptySource here
  @EnumSource(value = Sector.class, names = "UNKNOWN", mode = EnumSource.Mode.INCLUDE)
  void shouldNotPopulateSectorInApprovalViewWhenSectorIsNull(Sector sector) {
    var institutionId = randomOrganizationId();
    var candidate = setupCandidateWithInstitutionPoints(institutionId, sector, false);
    stubPublication(candidate);

    var document = generateIndexDocument(candidate);

    assertThat(approvalFor(document, institutionId).sector()).isNull();
  }
```

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/PublicationDetails.java
```diff
@@ -188,6 +188,12 @@ public List<UnverifiedNviCreatorDto> unverifiedCreators() {
         .toList();
   }
 
+  public Optional<Organization> findInstitution(URI institutionId) {
+    return topLevelOrganizations().stream()
+        .filter(organization -> institutionId.equals(organization.id()))
+        .findFirst();
```
brinxmat: First or any?

nvi-commons/src/test/java/no/sikt/nva/nvi/common/model/OrganizationTest.java
```diff
@@ -1,14 +1,53 @@
...
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
+import java.util.Collections;
+import java.util.List;
 import no.sikt.nva.nvi.common.client.model.Organization;
 import org.junit.jupiter.api.Test;
 
 class OrganizationTest {
 
+  @Test
```
brinxmat, line 17: There is a thought here that the organization tree for a given node is an ordered list of nodes, we should probably think about what the access pattern is and what data structure we need-and-when

brinxmat (approved):

index-handlers/src/main/java/no/sikt/nva/nvi/index/mapper/ContributorMapper.java
LarsV123: I considered this, but figured it was safe enough because "non-verified" contributors should only exist on unreported Candidates in non-closed Periods.

The utility is pretty marginal though, as the only useful bit of info we extract is the ORCID (and that is hopefully temporary). Remove it and skip supplemental fields for non-verified contributors?

index-handlers/src/main/java/no/sikt/nva/nvi/index/mapper/ContributorMapper.java
LarsV123: This would work here and now, but the next step is to add `role` and `orcid` fields to the persisted model. When that happens, the `extract*` methods first need to check both places like we do elsewhere. If we also migrate the historical records and persist the supplemental data, we can skip the enrichment entirely in this mapper.

index-handlers/src/main/java/no/sikt/nva/nvi/index/mapper/PublicationDetailsMapper.java
LarsV123: You're right. Will see if I do it in this PR or split it to a separate PR. Frankly, several of these models should probably be merged too.

index-handlers/src/main/java/no/sikt/nva/nvi/index/mapper/PublicationDetailsMapper.java
LarsV123: `findAny` should be fine and I don't think there can ever be more than one item in the collection anyway.

index-handlers/src/test/java/no/sikt/nva/nvi/index/IndexDocumentContentTest.java
LarsV123: Good tip, didn't know you could stack sources like that!

nvi-commons/src/main/java/no/sikt/nva/nvi/common/service/model/PublicationDetails.java
LarsV123: Any

nvi-commons/src/test/java/no/sikt/nva/nvi/common/model/OrganizationTest.java
LarsV123: Good point. The data is most likely an ordered list, but we don't rely on that assumption anywhere. We are only concerned with set membership, so I refactored this to be just that.

PR #788, opened 2026-07-14
refactor: Remove unused aggregations

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-51472

Removes search aggregations that are no longer in use and changes the default behavior from "all aggregations" to "no aggregations".


brinxmat (approved):
Possibly a breaking change if the aggregations are used by external users, but likely not

PR #791, opened 2026-07-21
[np-51491] feat: Persist ORCID for NVI creators

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-51491

Adds ORCID to the data we persist about NVI creators and sets the `role` as a static field (since it's the same for all persisted contributors). Expands the existing migration code so that we can backfill the `orcid` field based on the current publication state. Note that this adds data to the record that may not have existed at the time the candidate was reported.

Changes:

- [fix: Use correct field name when parsing ORCID](https://github.com/BIBSYSDEV/nva-nvi/pull/791/commits/a3d2ddf36a9101a7a6c8ea00410e75ad27d3b2c2)
- [fix: Use correct builder for DbUnverifiedCreator](https://github.com/BIBSYSDEV/nva-nvi/pull/791/commits/ff89212dd32955a62c9450107da3fb1b015321be)
- [feat: Persist ORCID for NVI creators](https://github.com/BIBSYSDEV/nva-nvi/pull/791/commits/87116fa9c69b1075715ad7be3e0836c0aacef833)

### DIFF
--- a/libs/publication-service/src/main/resources/nva_normalization.rq
+++ b/libs/publication-service/src/main/resources/nva_normalization.rq
@@ -139,7 +139,11 @@ WHERE {
         FILTER(?otherName > ?contributorName)
       }
     }
-    OPTIONAL { ?personId :orcId ?orcid }
+    # The ORCID is currently stored with the key "orcId" in expanded publications, but we
+    # accept both casing variants in case this is normalized to "orcid" upstream.
+    OPTIONAL { ?personId :orcId ?mixedCaseOrcid }
+    OPTIONAL { ?personId :orcid ?lowerCaseOrcid }
+    BIND(COALESCE(?mixedCaseOrcid, ?lowerCaseOrcid) AS ?orcid)
 
     # Find all valid affiliations for the contributor
     OPTIONAL {
--- a/libs/publication-service/src/main/resources/nva_normalization.rq
+++ b/libs/publication-service/src/main/resources/nva_normalization.rq
@@ -139,7 +139,7 @@ WHERE {
         FILTER(?otherName > ?contributorName)
       }
     }
-    OPTIONAL { ?personId :orcid ?orcid }
+    OPTIONAL { ?personId :orcId ?orcid }
 
     # Find all valid affiliations for the contributor
     OPTIONAL {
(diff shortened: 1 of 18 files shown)


libs/publication-service/src/main/resources/nva_normalization.rq
```diff
@@ -139,7 +139,7 @@ WHERE {
         FILTER(?otherName > ?contributorName)
       }
     }
-    OPTIONAL { ?personId :orcid ?orcid }
+    OPTIONAL { ?personId :orcId ?orcid }
```
brinxmat: I prefer the all lowercase variant, honestly. I don't think it is relevant to insert the "id" concept here, it is simply an IRI for us.

nvi-commons/src/testFixtures/java/no/sikt/nva/nvi/common/examples/ExampleContributors.java
```diff
@@ -44,6 +44,7 @@ public class ExampleContributors {
       ContributorDto.builder()
           .withId(URI.create("https://api.sandbox.nva.aws.unit.no/cristin/person/1685065"))
           .withName("Donald Duck")
+          .withOrcid(URI.create("https://fake.orcid.org/1234-5678-1234-5678"))
```
brinxmat, line 47: And this should then be `withOrcId`, which si horrible :D

brinxmat (commented):
I'd use `orcid` everywhere, including in the JSON.

libs/publication-service/src/main/resources/nva_normalization.rq
brinxmat: Lol, this is from the input? :D

brinxmat (approved):

nvi-commons/src/testFixtures/java/no/sikt/nva/nvi/common/examples/ExampleContributors.java
LarsV123: Should it?

libs/publication-service/src/main/resources/nva_normalization.rq
LarsV123: Yeah. I didn't bother matching the casing in our "output" here though, just fixed the input parsing.

PR #800, opened 2026-08-03
chore: Delete Cristin migration code

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-51480


brinxmat (approved):
Not bad separation here

